PR #4, opened 2026-01-26
feat: Add Phase 2 - VMBT/VMB creation and monitoring

shubham-pampattiwar:
## Summary

- Add `kubevirt.io/api v1.8.0-alpha.0` dependency for VirtualMachineBackup and VirtualMachineBackupTracker types
- Register KubeVirt backup types with the controller scheme
- Add RBAC permissions for `backup.kubevirt.io` resources and PVCs
- Implement Phase 2 controller logic:
  - `getVMReference()` - extract VM name/namespace from DataUpload annotations
    - Uses `kubevirt-datamover.io/vm-name` and `kubevirt-datamover.io/vm-namespace` annotations
  - `ensureTempPVC()` - create temporary PVC for backup output
  - `ensureVMBackupTracker()` - create/update VMBT for incremental backup tracking
  - `ensureVMBackup()` - create VMB CR referencing the VMBT
- Update `handleAccepted()` to orchestrate Phase 2 flow and monitor VMB status
  - Simplified VMB status detection: only checks `Done` condition
    - `Done: True` → Success, transition to Prepared
    - `Done: False` → Failure, transition to Failed
    - No `Done` condition → Still in progress, requeue
- Add comprehensive unit tests (61% coverage)
  - VMB status detection tests for all condition states
  - Helper function tests for getVMReference, updatePhase, etc.
- Fix PR automation workflow to support PRs from forks (`pull_request_target`)
- Fix deprecated `result.Requeue` usage (use `result.RequeueAfter` instead)

## Test plan

- [x] Unit tests pass locally (`make test`) - 61% coverage
- [x] Linter passes (`make lint`)
- [x] Deploy to cluster with KubeVirt
- [x] Create test DataUpload with `datamover: kubevirt` and VM annotations
- [x] Verify VMBT, VMB, and temp PVC are created correctly
- [x] Verify VMB status monitoring correctly detects `Done: True` → Prepared
- [x] Verify intermediate states (Progressing: False) don't cause false failures
- [x] Verify cleanup via owner references


internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -37,6 +45,27 @@ const (
...
+	AnnotationVMName = "kubevirt-datamover.io/vm-name"
+
+	// AnnotationVMNamespace is the annotation key for the VirtualMachine namespace
+	AnnotationVMNamespace = "kubevirt-datamover.io/vm-namespace"
+
+	// LabelDataUploadName is the label key for the DataUpload name
+	LabelDataUploadName = "velero.io/dataupload-name"
+
+	// LabelDataUploadUID is the label key for the DataUpload UID
+	LabelDataUploadUID = "velero.io/dataupload-uid"
```
mpryc: Is it worth to move this outside so plugin can take advantage, or should I replicate same on the plugin part? I am asking because I have and I am trying to make sure inconsistency in the future is not something we should worry about:

```go
// Annotation keys used by the kubevirt datamover plugin
const (
	// AnnDataUploadName is the annotation key for the DataUpload name on VirtualMachine
	AnnDataUploadName = "velero.io/data-upload-name"

	// AnnVMName is the annotation on DataUpload identifying the source VirtualMachine
	AnnVMName = "kubevirt.io/vm-name"

	// AnnVMNamespace is the annotation on DataUpload identifying the source VM namespace
	AnnVMNamespace = "kubevirt.io/vm-namespace"

	// AnnOperationID is the annotation for tracking async operation
	AnnOperationID = "kubevirt.io/operation-id"

	// AnnVolumePolicy is the annotation for volume policy on PVC
	AnnVolumePolicy = "velero.io/volume-policy"
)
```

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -113,36 +148,105 @@ func (r *KubeVirtDataUploadReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	for _, cond := range vmb.Status.Conditions {
+		if cond.Type == kubevirtbackupv1alpha1.ConditionDone {
+			if cond.Status == corev1.ConditionTrue {
+				// Success
+				logger.Info("VirtualMachineBackup completed",
+					"vmb", vmb.Name,
+					"type", vmb.Status.Type,
+					"checkpoint", vmb.Status.CheckpointName)
+
+				if err := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhasePrepared,
```
mpryc, line 243: Which of the phases will we have? On the plugin side I have for the progress based on the DU phase:

I see you have DataUploadPhaseAccepted, DataUploadPhasePrepared, DataUploadPhaseFailed so just a subset.

```go
	// Update progress based on DataUpload status
	switch dataUpload.Status.Phase {
	case velerov2alpha1.DataUploadPhaseNew:
		progress.Description = "DataUpload created, waiting for processing"
	case velerov2alpha1.DataUploadPhaseAccepted:
		progress.Description = "DataUpload accepted, preparing for backup"
	case velerov2alpha1.DataUploadPhasePrepared:
		progress.Description = "DataUpload prepared, starting data transfer"
	case velerov2alpha1.DataUploadPhaseInProgress:
		progress.Description = "Data transfer in progress"
		progress.NTotal = dataUpload.Status.Progress.TotalBytes
		progress.NCompleted = dataUpload.Status.Progress.BytesDone
	case velerov2alpha1.DataUploadPhaseCompleted:
		progress.Completed = true
		progress.Description = "Backup completed successfully"
		progress.NTotal = dataUpload.Status.Progress.TotalBytes
		progress.NCompleted = dataUpload.Status.Progress.TotalBytes
	case velerov2alpha1.DataUploadPhaseCanceled:
		progress.Completed = true
		progress.Err = "Backup was canceled"
		progress.Description = "Backup canceled"
	case velerov2alpha1.DataUploadPhaseCanceling:
		progress.Description = "Backup is being canceled"
	case velerov2alpha1.DataUploadPhaseFailed:
		progress.Completed = true
		progress.Err = dataUpload.Status.Message
		progress.Description = "Backup failed"
	}
```

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -251,3 +355,174 @@ func (r *KubeVirtDataUploadReconciler) filterKubeVirtDataMover() predicate.Predi
...
 }
+
+// getVMReference extracts the VirtualMachine name and namespace from DataUpload annotations
+func (r *KubeVirtDataUploadReconciler) getVMReference(du *velerov2alpha1.DataUpload) (string, string, error) {
+	annotations := du.GetAnnotations()
+	if annotations == nil {
+		return "", "", fmt.Errorf("DataUpload has no annotations")
+	}
+
+	vmName, ok := annotations[AnnotationVMName]
```
mpryc: Same here, should we have common shared annotations?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -37,6 +45,27 @@ const (
...
+	AnnotationVMNamespace = "kubevirt-datamover.io/vm-namespace"
+
+	// LabelDataUploadName is the label key for the DataUpload name
+	LabelDataUploadName = "velero.io/dataupload-name"
+
+	// LabelDataUploadUID is the label key for the DataUpload UID
+	LabelDataUploadUID = "velero.io/dataupload-uid"
+
+	// DefaultTempPVCSize is the default size for temporary backup PVC
+	DefaultTempPVCSize = "10Gi"
```
weshayutin, line 49: @shubham-pampattiwar  If the default is not used, where else is this set or calculated?
At a minimum I wonder if this size should go through a diferent code path based on whether we're doing a full backup or incremental right?

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Good call! I've created a shared `pkg/common/constants.go` package that the plugin can import:

```go
import "github.com/migtools/kubevirt-datamover-controller/pkg/common"
```

This gives us a single source of truth for:
- `common.AnnotationVMName` = `"kubevirt-datamover.io/vm-name"`
- `common.AnnotationVMNamespace` = `"kubevirt-datamover.io/vm-namespace"`
- `common.AnnotationOperationID` = `"kubevirt-datamover.io/operation-id"`
- `common.LabelDataUploadName` = `"velero.io/dataupload-name"`
- `common.LabelDataUploadUID` = `"velero.io/dataupload-uid"`
- `common.DataMoverKubeVirt` = `"kubevirt"`

Please update the plugin to use `kubevirt-datamover.io/` prefix (instead of `kubevirt.io/`) since we don't own that domain.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: The controller handles all the standard Velero DataUpload phases. Here's the complete flow:

**Phase Transitions (Controller):**
| From | To | Trigger |
|------|-----|---------|
| `New` | `Accepted` | VM annotations validated |
| `New` | `Failed` | Missing VM annotations |
| `Accepted` | `Prepared` | VMB completed (Done: True) |
| `Accepted` | `Failed` | VMB failed (Done: False) |
| `Prepared` | `InProgress` | Datamover pod started |
| `InProgress` | `Completed` | Data transfer complete (Phase 3) |
| `InProgress` | `Failed` | Transfer error |
| `InProgress` | `Canceling` | Cancel requested |
| `Canceling` | `Canceled` | Cleanup complete |

Your plugin progress descriptions look correct! The controller will transition through these phases and your plugin can report appropriate progress messages.

I've documented this in `pkg/common/constants.go` for reference.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Yes! Addressed in the same commit - the shared `pkg/common` package includes all the annotation and label constants.

The plugin should import:
```go
import "github.com/migtools/kubevirt-datamover-controller/pkg/common"

// Then use:
common.AnnotationVMName
common.AnnotationVMNamespace
common.AnnotationOperationID
```

This ensures both controller and plugin stay in sync.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -251,3 +355,174 @@ func (r *KubeVirtDataUploadReconciler) filterKubeVirtDataMover() predicate.Predi
...
+	if err == nil {
+		logger.V(1).Info("Temporary PVC already exists", "pvc", pvcName)
+		return existingPVC, nil
+	}
+
+	if !errors.IsNotFound(err) {
+		return nil, fmt.Errorf("failed to check for existing PVC: %w", err)
+	}
+
+	// Create new PVC
```
weshayutin, line 387: so perhaps rabbit.. here we could have two paths.   If the qcow2 snapshot is a full backup.. I think we MAY be able to calculate the required size of the pv based on the VM pv's themselves perhaps.   If the qcow2 snapshot is incremental, perhaps then we just go w/ the default size as specified by users?

sseago: Also, if the VMB fails because the PVC was too small, there will be a failure message indicating that. If that message indicates how far we fell short, we could have the controller create a new PVC, VMB and try again.

internal/controller/kubevirt_dataupload_controller.go
sseago: @weshayutin we need something a little more intelligent than simply full vs. incremental. As we saw with the etcd backups, incremental can be almost as large as full in some cases. With a VM, that's less likely if the OS/installed software portion of the storage is large in comparison to the data portion, but for example, if this is a VM running a large database, incremental backups could still be quite large. Actually, this is another reason we definitely need periodic forced full backups. A  year's worth of incremental backups may not fit on a PVC for restore.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Good point. Here's a proposed approach for smarter PVC sizing:

**Algorithm:**

1. **Check for user override** - Allow annotation `kubevirt-datamover.io/backup-pvc-size` to explicitly set size

2. **Query source VM's PVC sizes** - Look up the VM, iterate its volumes, sum the PVC sizes

3. **Check VMBT for existing checkpoint** - If `latestCheckpoint` exists → likely incremental

4. **Apply sizing logic:**
   - **Full backup** (no checkpoint): Use sum of source PVC sizes
   - **Incremental backup** (has checkpoint): Use 50% of total (or configurable fraction)
   - **Minimum floor**: Ensure at least 1Gi to handle edge cases

```go
func calculateBackupPVCSize(vm, vmbt, userOverride) Quantity {
    if userOverride != "" {
        return parse(userOverride)
    }
    
    totalSize := sum(vm.spec.volumes[*].pvc.size)
    
    if vmbt.status.latestCheckpoint != nil {
        // Incremental - use fraction
        return max(totalSize * 0.5, 1Gi)
    }
    
    // Full backup - use full size
    return totalSize
}
```

**Trade-offs:**
- This requires fetching the VM object (additional API call)
- 50% for incremental is a heuristic - database-heavy VMs may need more
- Could add retry logic later if VMB fails with "out of space"

Thoughts? Should I implement this in Phase 2 or track as follow-up?

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Great point about incremental != small. The 50% heuristic in my proposed algorithm is definitely just a starting point.

For database-heavy VMs, users could:
1. Use the annotation override to specify a larger size
2. Or we could expose a configurable "incremental ratio" (default 0.5, adjustable)

Re: periodic forced full backups - the VMBackup API already supports `forceFullBackup: true`. We could expose this as:
- A VMBT-level policy (e.g., "force full every N backups")
- Or honor a user annotation on the DataUpload

I'll create an issue to track "configurable incremental sizing + periodic full backup policy".

internal/controller/kubevirt_dataupload_controller.go
weshayutin: hrm.. not sure if this is accurate yet, but gemini has the following info re: qcow2 compression


Content Type | Typical Compression Ratio | Space Reduction
-- | -- | --
Fresh OS Install | 2:1 to 4:1 | 50% – 75%
Text/Logs/Code | 5:1 to 8:1 | 80% – 85%
Zero-filled Space | Near 100:1 | ~99%
Encrypted/Compressed Data | 1:1 | 0%

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -113,36 +134,105 @@ func (r *KubeVirtDataUploadReconciler) Reconcile(ctx context.Context, req ctrl.R
 func (r *KubeVirtDataUploadReconciler) handleNew(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload) (ctrl.Result, error) {
 	logger.Info("Handling New phase DataUpload")
 
-	// TODO Phase 2: Validate prerequisites
-	// - Check VM annotation exists
```
mpryc, line 117: common function pls that plugin is executing if possible, currently plugin has separate function.

shubham-pampattiwar:
## PVC Sizing Discussion - Tracked in Issue #5

Based on our call, we've decided to table the smart PVC sizing feature for now and track it separately.

**Current behavior (Phase 2):** Hardcoded 10Gi default

**Future enhancement:** Issue #5 captures the proposed solution including:
- Dynamic sizing based on source VM PVC sizes
- User override via annotation
- Configurable incremental ratio
- Retry logic on space exhaustion
- Periodic forced full backup policy

https://github.com/migtools/kubevirt-datamover-controller/issues/5

PR #6, opened 2026-01-28
Add shared GetPVCsForVM and SnapshotTypeCSI to pkg/common

mpryc:
Move PVC extraction logic and snapshot type constant to the shared common package so both the controller and kubevirt-datamover-plugin use the same implementation, ensuring consistency when processing VirtualMachine volumes.

### DIFF
--- a/pkg/common/validation.go
+++ b/pkg/common/validation.go
@@ -116,3 +116,37 @@ func ValidateVMForBackup(vm *kubevirtcorev1.VirtualMachine) error {
 
 	return nil
 }
+
+// GetVolumesForVm returns a list of PVC names associated with the VirtualMachine.
+// It extracts PVC names from:
+// - PersistentVolumeClaim volume sources
+// - DataVolume volume sources (which create PVCs with the same name)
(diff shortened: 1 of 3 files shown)


pkg/common/validation.go
```diff
@@ -116,3 +116,37 @@ func ValidateVMForBackup(vm *kubevirtcorev1.VirtualMachine) error {
 
 	return nil
 }
+
+// GetPVCsForVM returns a list of PVC names associated with the VirtualMachine.
+// It extracts PVC names from:
+// - PersistentVolumeClaim volume sources
+// - DataVolume volume sources (which create PVCs with the same name)
+// - MemoryDump volume sources
+func GetPVCsForVM(vm *kubevirtcorev1.VirtualMachine) []string {
```
shubham-pampattiwar: Should this just be `GetVolumesForVm` ?

mpryc: Updated.

shubham-pampattiwar (approved):

PR #10, opened 2026-02-05
feat: Add Phase 3 - Datamover pod for object storage upload

### LINKED ISSUE #2064: kubevirt-dm: DataMover Pod/Uploader Actions
Description:
Create/update the appropriate BSL manifest/checkpoint files for the VM and upload these and upload the qcow2 files.
Update the DataUpload status appropriately (including the total/incremental size)

Subtasks:

DOD:

shubham-pampattiwar:
## Summary

This implements Phase 3 of the kubevirt-datamover-controller: the datamover pod that uploads qcow2 files from the temporary PVC (created in Phase 2) to object storage (BSL).

- Datamover pod with `upload` subcommand (same image as controller)
- S3 object storage integration with AWS SDK
- PV rebinding from VM namespace to OADP namespace using Velero's Patch pattern
- SELinux access via `spc_t` type (follows Velero's CSI snapshot approach)
- S3 storage structure aligned with approved design document
- Configurable datamover image via `DATAMOVER_IMAGE` env var (synced with controller image via kustomize)

Fixes: https://github.com/openshift/oadp-operator/issues/2064 
Parent issue: https://github.com/openshift/oadp-operator/issues/2063

## S3 Storage Layout

```
<prefix>-kubevirt-datamover/
├── checkpoints/<ns>/<vm>/
│   ├── <checkpoint-id>/<vmb>-<disk>.qcow2
│   └── index.json           # Per-VM checkpoint index
└── manifests/<backup>/
    ├── index.json           # Per-backup manifest
    └── <ns>-<vm>.json       # Per-VM backup manifest with checkpoint chain
```

## Files Added

| File | Description |
|------|-------------|
| `cmd/uploader/main.go` | Uploader entrypoint |
| `pkg/uploader/types.go` | S3 metadata types with documentation |
| `pkg/uploader/run.go` | Upload workflow |
| `pkg/uploader/objectstore.go` | S3 ObjectStore implementation |
| `internal/controller/pod_builder.go` | Datamover pod specification |
| `internal/controller/pv_rebind.go` | PV rebinding using Patch operations |

### Prerequisites

1. OpenShift cluster with:
   - OpenShift Virtualization installed
   - ODF/Ceph storage (or any CSI storage with snapshot support)
   - A running VirtualMachine

2. OADP installed with a BackupStorageLocation configured:
   ```bash
   kubectl get backupstoragelocation -n openshift-adp
   # Note the BSL name (e.g., sample-dpa-1)
   ```

### Step 1: Build and Deploy Controller

```bash
# Clone and build
git clone https://github.com/migtools/kubevirt-datamover-controller.git
cd kubevirt-datamover-controller
git checkout phase3-datamover-pod

# Build image for your architecture (amd64 for OCP)
docker build --platform linux/amd64 -t <your-registry>/kubevirt-datamover-controller:latest .

# Push to registry
docker push <your-registry>/kubevirt-datamover-controller:latest

# Deploy to cluster
make deploy IMG=<your-registry>/kubevirt-datamover-controller:latest
```

### Step 2: Verify Controller is Running

```bash
kubectl get pods -n openshift-adp -l control-plane=controller-manager

# Check DATAMOVER_IMAGE env var is set correctly
kubectl get deployment -n openshift-adp kubevirt-datamover-controller-manager \
  -o jsonpath='{.spec.template.spec.containers[0].env}' | jq .
```

### Step 3: Create Test DataUpload

Replace `<bsl-name>`, `<vm-name>`, and `<vm-namespace>` with your values:

```yaml
apiVersion: velero.io/v2alpha1
kind: DataUpload
metadata:
  name: test-phase3-upload
  namespace: openshift-adp
  labels:
    velero.io/backup-name: test-backup
  annotations:
    kubevirt-datamover.io/vm-name: <vm-name>
    kubevirt-datamover.io/vm-namespace: <vm-namespace>
spec:
  snapshotType: CSI
  csiSnapshot:
    storageClass: ocs-storagecluster-ceph-rbd
    volumeSnapshot: placeholder-not-used
  sourcePVC: <any-pvc-name>
  sourceNamespace: <vm-namespace>
  datamover: kubevirt
  backupStorageLocation: <bsl-name>
  operationTimeout: 10m
```

Apply:
```bash
kubectl apply -f dataupload.yaml
```

### Step 4: Monitor Progress

```bash
# Watch DataUpload phase transitions
kubectl get dataupload -n openshift-adp -w

# Check controller logs
kubectl logs -n openshift-adp deployment/kubevirt-datamover-controller-manager -f

# Check VMB status (in VM namespace)
kubectl get vmb -n <vm-namespace>

# Check datamover pod (when InProgress)
kubectl get pods -n openshift-adp -l kubevirt-datamover.io/pod-type=datamover
```

Expected phase transitions: `New → Accepted → Prepared → InProgress → Completed`

### Step 5: Verify S3 Upload

```bash
# Get BSL bucket and prefix
kubectl get backupstoragelocation -n openshift-adp <bsl-name> \
  -o jsonpath='{.spec.objectStorage}' | jq .

# List uploaded files
aws s3 ls s3://<bucket>/<prefix>-kubevirt-datamover/ --recursive

# Check VM index
aws s3 cp s3://<bucket>/<prefix>-kubevirt-datamover/checkpoints/<vm-namespace>/<vm-name>/index.json - | jq .

# Check backup manifest
aws s3 cp s3://<bucket>/<prefix>-kubevirt-datamover/manifests/test-backup/index.json - | jq .
```

### Step 6: Verify Cleanup

After completion, verify resources are cleaned up:
```bash
# Datamover pod should be deleted
kubectl get pods -n openshift-adp -l kubevirt-datamover.io/pod-type=datamover

# Rebound PVC should be deleted
kubectl get pvc -n openshift-adp -l velero.io/dataupload-name=test-phase3-upload

# PV should be deleted
kubectl get pv -l velero.io/dataupload-name=test-phase3-upload
```

### Cleanup Test Resources

```bash
kubectl delete dataupload -n openshift-adp test-phase3-upload
kubectl delete vmb -n <vm-namespace> --all
```

## Test plan

- [x] Single VM backup completes successfully
- [x] Incremental backup chain works correctly (parent links)
- [x] S3 storage structure matches design
- [x] Cleanup on success (pod, PVC, PV deleted)
- [x] Cleanup on failure re-enabled
- [x] Unit tests for uploader functions 
- [x] DATAMOVER_IMAGE env var properly synced with controller image


cmd/main.go
```diff
@@ -88,6 +105,12 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 3,
 		"Maximum number of concurrent reconciles for the KubeVirt DataUpload controller")
+	flag.StringVar(&datamoverImage, "datamover-image", "",
```
mpryc: Do we have some default here? I assume it's always referenced by hash not tag, which causes `IfNotPresent` work as expected pooling always required image?

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: quay.io/spampatt/kubevirt-datamover-controller
```
mpryc: Don't we want to take from the konveyor ?

pkg/uploader/objectstore.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	return credentials.NewStaticCredentialsProvider(accessKeyID, secretAccessKey, ""), nil
+}
+
+// InitObjectStore creates an ObjectStore based on the provider type.
+// Returns a velero.ObjectStore implementation.
+func InitObjectStore(cfg *UploaderConfig) (velero.ObjectStore, error) {
+	switch strings.ToLower(cfg.BSLProvider) {
+	case "aws", "":
```
mpryc: Is the "" as well in velero code? we have default but "" seems like default here?

pkg/uploader/run.go
```diff
@@ -0,0 +1,409 @@
...
+		config.VMNamespace, config.VMName, config.CheckpointName, config.BackupType)
+
+	// Initialize object store
+	storeInterface, err := InitObjectStore(config)
+	if err != nil {
+		return fmt.Errorf("failed to initialize object store: %w", err)
+	}
+
+	// Type assert to get the concrete type with convenience methods
+	store, ok := storeInterface.(*S3ObjectStore)
```
mpryc: Is the S3ObjectStore also an object for GCP/Azure/other.. ?

pkg/uploader/objectstore.go
```diff
@@ -0,0 +1,331 @@
...
+		return NewS3ObjectStore(
+			cfg.BSLBucket,
+			cfg.BSLPrefix,
+			cfg.BSLRegion,
+			cfg.CredentialsFile,
+		)
+	case "gcp":
+		return nil, fmt.Errorf("gcp object store not yet implemented")
```
mpryc, line 340: We need to include this as issue / TODO so we don't forget...

mpryc: Also is the S3ObjectStore a common object for aws/gcp/azure/default ? it's being used in run.go / added comment there.

pkg/uploader/objectstore.go
```diff
@@ -0,0 +1,331 @@
...
+// ObjectExists checks if an object exists in S3.
+// Implements velero.ObjectStore.ObjectExists().
+func (s *S3ObjectStore) ObjectExists(bucket, key string) (bool, error) {
+	_, err := s.client.HeadObject(context.Background(), &s3.HeadObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(s.fullKey(key)),
+	})
+	if err != nil {
+		// Check if it's a not found error
+		if strings.Contains(err.Error(), "NotFound") || strings.Contains(err.Error(), "404") {
```
mpryc: Why not use aws sdk errors instead of those string comparisons ? 
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3/types#NotFound

pkg/uploader/objectstore.go
```diff
@@ -0,0 +1,331 @@
...
+	// Load credentials from file if specified
+	if credentialsFile != "" {
+		creds, err := loadAWSCredentialsFromFile(credentialsFile)
+		if err != nil {
+			return fmt.Errorf("failed to load credentials: %w", err)
+		}
+		opts = append(opts, config.WithCredentialsProvider(creds))
+	}
+
+	cfg, err := config.LoadDefaultConfig(ctx, opts...)
```
mpryc, line 113: We should consider including retry to operations to limit transient errors:
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws/retry

```go
"github.com/aws/aws-sdk-go-v2/aws/retry"

[...]

    // Add retry configuration
    opts = append(opts, config.WithRetryer(func() aws.Retryer {
        return retry.NewStandard(func(o *retry.StandardOptions) {
            o.MaxAttempts = 3
        })
    }))
  

  cfg, err := config.LoadDefaultConfig(ctx, opts...)
```

internal/controller/pv_rebind.go
```diff
@@ -0,0 +1,324 @@
...
+			if errors.IsNotFound(err) {
+				logger.V(1).Info("PV already deleted", "pv", pvName)
+				return nil
+			}
+			return fmt.Errorf("failed to get PV %s: %w", pvName, err)
+		}
+
+		// Set reclaim policy to Delete to ensure underlying storage is cleaned up
+		if err := r.patchPVReclaimPolicy(ctx, pv, corev1.PersistentVolumeReclaimDelete); err != nil {
+			logger.Error(err, "Failed to set PV reclaim policy to Delete", "pv", pvName)
```
mpryc: Should there be a retry? If the patch fails here and we remove the pv it could lead to storage being not deleted on the underlying storage system - leakage of storage occupied as we delete pv, so cluster does not track it anymore. Maybe at this stage we should not delete PV ?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -262,13 +273,141 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
+		if err := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhaseFailed, fmt.Sprintf("Missing VM reference: %v", err)); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Datamover pod runs in OADP namespace (where credentials are accessible)
+	podNamespace := r.OADPNamespace
+	if podNamespace == "" {
+		podNamespace = du.Namespace
```
mpryc: Should it fail if it's not OADP NS?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -262,13 +273,141 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
+	}
+
+	// Check if datamover pod already exists (idempotency)
+	podName := getDatamoverPodName(du)
+	existingPod := &corev1.Pod{}
+	err = r.Get(ctx, types.NamespacedName{Name: podName, Namespace: podNamespace}, existingPod)
+	if err == nil {
+		// Pod exists, transition to InProgress and monitor
+		logger.Info("Datamover pod already exists, transitioning to InProgress", "pod", podName)
+		if err := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhaseInProgress, "Datamover pod running"); err != nil {
```
mpryc, line 300: is updatePhase taking care of reconcile event in which it was already in progress? so skips update?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -262,13 +273,141 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
+
+		logger.Info("Rebinding PV from VM namespace to OADP namespace",
+			"sourcePVC", sourcePVCName,
+			"sourceNamespace", vmRef.Namespace,
+			"targetNamespace", podNamespace)
+
+		rebindResult, err := r.rebindPVToNamespace(ctx, logger, sourcePVCName, vmRef.Namespace, podNamespace, du.Name)
+		if err != nil {
+			logger.Error(err, "Failed to rebind PV to OADP namespace")
+			if err := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhaseFailed, fmt.Sprintf("Failed to rebind PV: %v", err)); err != nil {
```
mpryc, line 354: What we should do now? retry?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -262,13 +273,141 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
+			"newPVC", rebindResult.NewPVCName,
+			"pv", rebindResult.PVName)
+
+		reboundPVCName = rebindResult.NewPVCName
+	} else {
+		logger.Info("PV already rebound to OADP namespace", "pvc", reboundPVCName)
+	}
+
+	// Get BackupStorageLocation
+	bsl, err := r.getBackupStorageLocation(ctx, du)
```
mpryc: I think those checks should happen prior to PV rebinding. much less operations in case of failure to get bsl.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -283,30 +422,103 @@ func (r *KubeVirtDataUploadReconciler) handlePrepared(ctx context.Context, logge
...
 	logger.Info("Handling InProgress phase DataUpload")
 
-	// TODO Phase 3: Monitor datamover pod
-	// - Check pod status
-	// - On success: update index.json, create manifests, transition to Completed
-	// - On failure: transition to Failed
+	// Datamover pod runs in OADP namespace
+	podNamespace := r.OADPNamespace
+	if podNamespace == "" {
+		podNamespace = du.Namespace
```
mpryc: Sounds like could be getOrCreatePod ?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -283,30 +422,103 @@ func (r *KubeVirtDataUploadReconciler) handlePrepared(ctx context.Context, logge
...
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+
+	case corev1.PodFailed:
+		failureMessage := extractPodFailureMessage(pod)
+		logger.Error(nil, "Datamover pod failed", "pod", podName, "message", failureMessage)
+
+		// Cleanup resources on failure
+		r.cleanupDatamoverResources(ctx, logger, du, podNamespace)
```
mpryc: on a failure cleanup? I see we have log extraceted, is this sufficient?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -283,30 +422,103 @@ func (r *KubeVirtDataUploadReconciler) handlePrepared(ctx context.Context, logge
...
 
-	// TODO: Cancel in-progress operations
-	// - Delete datamover pod if running
-	// - Delete VMB if exists
-	// - Delete temporary PVC
-	// - Transition to Canceled
+	// Datamover pod runs in OADP namespace
+	podNamespace := r.OADPNamespace
+	if podNamespace == "" {
+		podNamespace = du.Namespace
```
mpryc: Similar to previous, seems like area to improve. common func at least or fail if not oadp ns from oadp controller.

internal/controller/pod_builder.go
```diff
@@ -0,0 +1,171 @@
...
+	readOnlyRootFilesystem := true
+
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      config.Name,
+			Namespace: config.Namespace,
+			Labels:    labels,
+		},
+		Spec: corev1.PodSpec{
+			// Use velero service account which has the SCC needed for root access
```
mpryc, line 111: I wonder why we need UID 0 access, this is series of patch files created by kubevirt, so isn't that we need to obtain scc from kubevirt ?

internal/controller/pod_builder.go
```diff
@@ -0,0 +1,171 @@
...
+				},
+				{
+					Name: "cloud-credentials",
+					VolumeSource: corev1.VolumeSource{
+						Secret: &corev1.SecretVolumeSource{
+							SecretName: config.CredentialSecretName,
+							Items: []corev1.KeyToPath{
+								{
+									Key:  config.CredentialSecretKey,
+									Path: "cloud",
```
mpryc: always?

internal/controller/pv_rebind.go
```diff
@@ -0,0 +1,324 @@
...
+	if err := r.Get(ctx, types.NamespacedName{Name: pvName}, pv); err != nil {
+		return nil, fmt.Errorf("failed to get PV %s: %w", pvName, err)
+	}
+
+	logger.Info("Found PV bound to source PVC", "pv", pvName, "sourcePVC", sourcePVCName)
+
+	// Step 2: Set PV reclaim policy to Retain using Patch
+	originalReclaimPolicy := pv.Spec.PersistentVolumeReclaimPolicy
+	if pv.Spec.PersistentVolumeReclaimPolicy != corev1.PersistentVolumeReclaimRetain {
+		if err := r.patchPVReclaimPolicy(ctx, pv, corev1.PersistentVolumeReclaimRetain); err != nil {
```
mpryc, line 110: Should there be a retry on patch error?

internal/controller/pv_rebind.go
```diff
@@ -0,0 +1,324 @@
...
+	logger.Info("Created new PVC in target namespace", "pvc", newPVCName, "namespace", targetNamespace)
+
+	// Step 5: Reset PV binding using Patch (like Velero's ResetPVBinding)
+	// Re-fetch PV to get latest version
+	if err := r.Get(ctx, types.NamespacedName{Name: pvName}, pv); err != nil {
+		return nil, fmt.Errorf("failed to re-fetch PV %s: %w", pvName, err)
+	}
+
+	labels := map[string]string{labelKey: labelValue}
+	if err := r.patchPVBinding(ctx, pv, newPVC, labels); err != nil {
```
mpryc, line 171: Similar here patch&retry

kaovilai:
@coderabbitai full review

cmd/main.go
shubham-pampattiwar: Added a default value using constant `common.DefaultDatamoverImage` set to `quay.io/konveyor/kubevirt-datamover-controller:latest`. This image will be built via CI, and in production the datamover image will be set by the OADP operator during installation, which will override this default with the appropriate image tag/digest.

config/manager/kustomization.yaml
shubham-pampattiwar: Updated to use `quay.io/konveyor/kubevirt-datamover-controller:latest`

pkg/uploader/objectstore.go
shubham-pampattiwar: Changed to return an error if BSL provider is empty: `"BSL provider is required but was empty"`. This makes configuration explicit rather than silently defaulting to AWS when provider is not set.

pkg/uploader/run.go
shubham-pampattiwar: Added TODO comment: `// TODO: Refactor to use a common interface when GCP/Azure are added (issue #11)`. Currently only S3 is implemented; when GCP/Azure support is added, we'll create a common interface with provider-specific implementations.

pkg/uploader/objectstore.go
shubham-pampattiwar: Created GitHub issue #11 to track GCP/Azure object store support. Added TODO comments in the code referencing this issue. S3ObjectStore is currently AWS-specific; we'll create a common interface when adding other providers.

pkg/uploader/objectstore.go
shubham-pampattiwar: Updated to use AWS SDK error types: `errors.As(err, &s3types.NotFound{})` and `errors.As(err, &s3types.NoSuchKey{})` instead of string comparisons.

pkg/uploader/objectstore.go
shubham-pampattiwar: Added retry configuration using AWS SDK's built-in retry mechanism: `retry.NewStandard` with `MaxAttempts = 3`. This handles transient errors like network timeouts automatically.

internal/controller/pv_rebind.go
shubham-pampattiwar: Added retry logic with 3 attempts for the reclaim policy patch. If the patch fails after all retries, we now return an error without deleting the PV to prevent storage leakage. The error message explicitly states: `"cannot delete PV: failed to set reclaim policy to Delete (would cause storage leakage)"`.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: In practice, DataUpload will always be in the OADP namespace since the BIA plugin creates it there. Added `getPodNamespace()` helper that uses `OADPNamespace` if configured, with fallback to DataUpload namespace. This matches how Velero node-agent operates.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Yes, added idempotency check to `updatePhase`: if `du.Status.Phase == phase`, it returns early without updating. This prevents unnecessary API calls and reconcile loops when the phase is already correct.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Added comment explaining the decision: PV rebinding involves destructive operations (source PVC deletion). If it fails mid-way, the system may be in an inconsistent state where retrying isn't safe. We fail fast to allow manual investigation rather than risk data corruption.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Moved BSL and VMB validation checks before PV rebinding. Now we validate that BSL credentials and VMB status are available before performing any destructive operations.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Considered this, but the current flow is already structured: we check for existing pod early in handlePrepared (idempotency check), then create if needed at the end. The logic is linear and clear without an additional abstraction layer.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Good point. Changed to skip cleanup on failure to preserve resources for debugging. Pod, rebound PVC/PV are now preserved when datamover pod fails, allowing investigation. Manual cleanup is required after debugging.

internal/controller/kubevirt_dataupload_controller.go
shubham-pampattiwar: Created `getPodNamespace()` helper function to consolidate this logic. It uses `OADPNamespace` if configured, with fallback to DataUpload namespace.

internal/controller/pod_builder.go
shubham-pampattiwar: Expanded the comment to explain: qcow2 files on the PVC may be owned by qemu user (UID 107) or vary by storage provider. Running as root ensures we can read the files regardless of ownership. We also use `spc_t` SELinux type to bypass SELinux label checking for cross-namespace volume access (the PV is rebound from VM namespace to OADP namespace). This follows Velero's pattern for CSI snapshot pods on OpenShift. The `velero` service account provides the necessary SCC.

internal/controller/pod_builder.go
shubham-pampattiwar: Added comment explaining: The BSL credentials secret key is mounted to a fixed path `/credentials/cloud` (matching `uploader.DefaultCredentialsPath`). This is simpler than Velero's dynamic path approach since we control both the pod builder and uploader code. If BSL has a custom credential key, it's still mounted correctly via the `KeyToPath` mapping.

internal/controller/pv_rebind.go
shubham-pampattiwar: Added retry logic with 3 attempts and 1 second interval between retries. On each retry, we re-fetch the PV to get the latest resource version to avoid conflicts.

internal/controller/pv_rebind.go
shubham-pampattiwar: Added retry logic with 3 attempts and 1 second interval between retries. On each retry, we re-fetch the PV to get the latest resource version to avoid conflicts.

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
...
+  newTag: latest
```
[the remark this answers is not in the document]
shubham-pampattiwar: Updated to use `quay.io/konveyor/kubevirt-datamover-controller:latest`. In production, the specific image tag will be set by CI/OADP operator.

internal/controller/pv_rebind.go
```diff
@@ -0,0 +1,324 @@
...
+				},
+			},
+			StorageClassName: sourcePVC.Spec.StorageClassName,
+			VolumeMode:       sourcePVC.Spec.VolumeMode,
+		},
+	}
+
+	if err := r.Create(ctx, newPVC); err != nil {
+		return nil, fmt.Errorf("failed to create new PVC %s/%s: %w", targetNamespace, newPVCName, err)
+	}
```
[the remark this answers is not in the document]
shubham-pampattiwar: Addressed - now handling `AlreadyExists` error as success instead of failure. This makes the operation idempotent for reconcile retries.

internal/controller/pv_rebind.go
```diff
@@ -0,0 +1,324 @@
...
+	})
+}
+
+// waitForPVCBound waits for a PVC to be bound
+func (r *KubeVirtDataUploadReconciler) waitForPVCBound(ctx context.Context, pvcName, namespace string) error {
+	return wait.PollUntilContextTimeout(ctx, PVRebindPollInterval, PVRebindTimeout, true, func(ctx context.Context) (bool, error) {
+		pvc := &corev1.PersistentVolumeClaim{}
+		if err := r.Get(ctx, types.NamespacedName{Name: pvcName, Namespace: namespace}, pvc); err != nil {
+			return false, err
+		}
```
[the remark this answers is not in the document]
shubham-pampattiwar: Addressed - now returning a descriptive error when PVC is deleted while waiting: `"PVC was deleted while waiting for it to bind"`

internal/controller/pv_rebind.go
```diff
@@ -0,0 +1,324 @@
...
+) error {
+	// Get the PVC to find the PV name
+	pvc := &corev1.PersistentVolumeClaim{}
+	err := r.Get(ctx, types.NamespacedName{Name: pvcName, Namespace: pvcNamespace}, pvc)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.V(1).Info("Rebound PVC already deleted", "pvc", pvcName)
+			return nil
+		}
+		return fmt.Errorf("failed to get rebound PVC %s/%s: %w", pvcNamespace, pvcName, err)
```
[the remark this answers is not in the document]
shubham-pampattiwar: If the PVC is already gone when cleanup runs, we return early without knowing which PV to clean up (we get PV name from `pvc.Spec.VolumeName`). In this edge case, the PV may need manual cleanup. Tracking PV name separately would be a future enhancement.

pkg/uploader/run_test.go
```diff
@@ -0,0 +1,659 @@
...
+					vmIndex.Checkpoints = append(vmIndex.Checkpoints, checkpoint)
+				}
+
+				indexData, _ := json.Marshal(vmIndex)
+				_ = store.PutObjectBytes(indexPath, indexData)
+
+				tt.validateResult(t, store)
+			}
+		})
+	}
```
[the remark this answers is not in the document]
shubham-pampattiwar: This was analysis by the automated reviewer. The test file has been updated with additional test cases for checkpoint name validation and invalid backup type.

pkg/uploader/run.go
```diff
@@ -0,0 +1,409 @@
...
+	// Validate required fields
+	if config.BSLBucket == "" {
+		return nil, fmt.Errorf("%s is required", EnvBSLBucket)
+	}
+	if config.VMName == "" {
+		return nil, fmt.Errorf("%s is required", EnvVMName)
+	}
+	if config.VMNamespace == "" {
+		return nil, fmt.Errorf("%s is required", EnvVMNamespace)
+	}
```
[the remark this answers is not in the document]
shubham-pampattiwar: Addressed - added validation for required `CheckpointName` field and `BackupType` (must be 'full' or 'incremental').

pkg/uploader/run.go
```diff
@@ -0,0 +1,409 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to open file %s: %w", path, err)
+		}
+		defer func() { _ = file.Close() }()
+
+		fmt.Printf("Uploading %s (%d bytes) to %s\n", d.Name(), info.Size(), objectPath)
+
+		// Upload file using convenience method
+		if err := store.PutObjectWithBucket(objectPath, file); err != nil {
+			return fmt.Errorf("failed to upload %s: %w", path, err)
```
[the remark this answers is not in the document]
shubham-pampattiwar: Addressed - files are now closed immediately after upload instead of using defer, preventing resource exhaustion with many files.

pkg/uploader/run.go
```diff
@@ -0,0 +1,409 @@
...
+			vmIndex = VMIndex{}
+		}
+	} else {
+		// Index doesn't exist, create new
+		vmIndex = VMIndex{
+			VMName:      config.VMName,
+			Namespace:   config.VMNamespace,
+			Checkpoints: []CheckpointEntry{},
+		}
+	}
```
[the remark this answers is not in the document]
shubham-pampattiwar: Addressed - now using `ObjectExists()` to check if index exists before attempting to read. This distinguishes 'not found' from actual read errors (network issues, etc.).

pkg/uploader/run.go
```diff
@@ -0,0 +1,409 @@
...
+
+	var vmIndex VMIndex
+	if err := json.Unmarshal(data, &vmIndex); err != nil {
+		return fmt.Errorf("failed to parse VM index: %w", err)
+	}
+
+	// Build checkpoint chain (for restore)
+	chain := buildCheckpointChain(vmIndex.Checkpoints, config.CheckpointName)
+
```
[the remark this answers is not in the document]
shubham-pampattiwar: Addressed - now returning an error if checkpoint chain is empty: `"failed to build checkpoint chain: checkpoint not found in VM index"`

test/e2e/e2e_test.go
```diff
@@ -34,7 +34,7 @@ import (
 )
 
 // namespace where the project is deployed in
-const namespace = "kubevirt-datamover-system"
+const namespace = "openshift-adp"
```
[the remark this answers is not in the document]
shubham-pampattiwar: Addressed - namespace creation now ignores 'already exists' error for pre-existing namespaces like openshift-adp.

internal/controller/pv_rebind.go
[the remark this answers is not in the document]
shubham-pampattiwar: Fixed in commit 77d6274. Changed to treat NotFound as transient cache race (return `false, nil` to continue polling) instead of returning an error. This handles the case where a cached client temporarily doesn't see the PVC after creation.

internal/controller/pv_rebind.go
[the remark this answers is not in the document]
shubham-pampattiwar: Fixed in commit 77d6274. Added `dataUploadName` parameter to `cleanupReboundPVCAndPV`. When the PVC is already deleted, we now find the PV by the `LabelDataUploadName` label (which was set during rebind) and still delete it, preventing storage leakage.

pkg/uploader/run_test.go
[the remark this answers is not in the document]
shubham-pampattiwar: Fixed in commit 77d6274. Refactored to use `velero.ObjectStore` interface instead of concrete `*S3ObjectStore` type. Added helper functions `putObjectBytes` and `getObjectBytes` that work with the interface. Tests now call the real `updateVMIndex` function with `MockObjectStore`, ensuring proper test coverage of the actual implementation logic including case-insensitive backup type comparison, timestamp fields, etc.

weshayutin (dismissed):
looking forward to trying this out

sseago (dismissed):

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -321,6 +536,14 @@ func (r *KubeVirtDataUploadReconciler) handleCanceling(ctx context.Context, logg
 func (r *KubeVirtDataUploadReconciler) updatePhase(ctx context.Context, du *velerov2alpha1.DataUpload, phase velerov2alpha1.DataUploadPhase, message string) error {
 	logger := log.FromContext(ctx)
 
+	// Skip update if already at target phase (idempotency)
+	if du.Status.Phase == phase {
+		logger.V(1).Info("DataUpload already at target phase, skipping update",
+			"dataUpload", du.Name,
+			"phase", phase)
+		return nil
+	}
```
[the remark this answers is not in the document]
shubham-pampattiwar: Fixed. The `updatePhase` function now also checks if the message is the same before skipping the update.

pkg/uploader/run.go
```diff
@@ -0,0 +1,447 @@
...
+	vmManifestData, err := json.MarshalIndent(vmBackupManifest, "", "  ")
+	if err != nil {
+		return fmt.Errorf("failed to marshal VM backup manifest: %w", err)
+	}
+
+	if err := store.PutObjectBytes(vmManifestPath, vmManifestData); err != nil {
+		return fmt.Errorf("failed to write VM backup manifest: %w", err)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
shubham-pampattiwar: Added a warning log if the checkpoint chain does not start with a full backup. We log a warning rather than fail because: (1) this could be the first backup marked as incremental by mistake, (2) the restore logic can still attempt to restore from whatever checkpoints are available. The warning alerts operators to investigate the chain integrity.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -262,13 +273,144 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
+		if err != nil {
+			// Fail without retry: PV rebind is a multi-step operation (delete PVC, patch PV, create new PVC).
+			// If it fails partway through, automatic retries could leave resources in an inconsistent state.
+			// Failing allows the user to investigate and take corrective action.
+			logger.Error(err, "Failed to rebind PV to OADP namespace")
+			if err := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhaseFailed, fmt.Sprintf("Failed to rebind PV: %v", err)); err != nil {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, nil
+		}
```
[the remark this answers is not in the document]
shubham-pampattiwar: Keeping the current check (`ClaimBound`) as a safety measure. The "Fail without retry" pattern already handles partial rebind states safely - if the rebound PVC exists but is Pending, the rebind attempt will fail at step 1 (source PVC gone), and the DataUpload transitions to Failed. This allows the operator to investigate rather than potentially making things worse with automatic retries. The current behavior is intentional and safe.

mpryc (commented):
/lgtm

Thanks @shubham-pampattiwar

weshayutin (approved):
/LGTM

mpryc (approved):

Issue #12, opened 2026-02-13
Phase 5: Complete cleanup handling and VMB/VMBT S3 archival


shubham-pampattiwar:
## Summary

Implement complete cleanup handling for DataUpload completion and failure cases, including archiving VMB/VMBT CRs to S3.

## Parent Issue
https://github.com/openshift/oadp-operator/issues/2063

## Related Issue
https://github.com/openshift/oadp-operator/issues/2061 (DeleteItemAction for S3 cleanup)

## Current Behavior

| Resource | On Success | On Failure |
|----------|------------|------------|
| Datamover Pod | ✅ Deleted | ❌ Preserved |
| Rebound PVC/PV | ✅ Deleted | ❌ Preserved |
| VMB | ❌ NOT deleted | ❌ Preserved |
| VMBT | ❌ NOT deleted | ❌ Preserved |
| VMB/VMBT in S3 | ❌ NOT uploaded | ❌ NOT uploaded |

## Proposed Behavior

| Resource | On Success | On Failure |
|----------|------------|------------|
| Datamover Pod | ✅ Delete |   ✅ Delete (enhance via --debug flag etc) |
| Rebound PVC/PV | ✅ Delete | ✅ Delete (enhance via --debug flag etc) |
| VMB | ✅ Delete (after S3 upload) | ✅ Delete (new for each DU/Backup operation)| 
| VMBT | ✅ Delete (after S3 upload) | ✅ Delete (Recreate from S3 and patch with the correct checkpoint per DU/Backup operation) |
| VMB/VMBT in S3 | ✅ Upload as JSON | ❌ Skip (incomplete backup) |

Note(modifications for phase 4): 
- Before creating the VMB object check for correct checkpoint in status of VMBT created by us (recreated from s3), We recreate/create VMBT from s3 and patch the VMBT status with the correct checkpoint and the make sure the checkpoint.
- If VMBT already exists in the cluster with the same name and all metadata, we delete it and re-create it from the s3 json, we treat s3 as source of truth at all times.

### 1. Upload VMB/VMBT to S3 (on success) (phase 4 task as per discussion)
Store in our custom checkpoint path (similar to how Velero stores resources as JSON):
- `checkpoints/<ns>/<vm>/<checkpoint-id>/vmb.json` - VMB CR as JSON
- `checkpoints/<ns>/<vm>/<checkpoint-id>/vmbt.json` - VMBT CR as JSON (updated per checkpoint)
[...]

PR #13, opened 2026-02-16
feat: Add Phase 4 - BSL checkpoint validation for incremental backup integrity

### LINKED ISSUE #2064: kubevirt-dm: DataMover Pod/Uploader Actions

mpryc:
## Summary

This implements Phase 4 of the kubevirt-datamover-controller: BSL checkpoint validation for incremental backup chain integrity.

- BSL checkpoint lookup validates existing checkpoint chains in S3 before creating a new backup
- Broken chain detection: missing qcow2 files or missing index.json triggers automatic fallback to full backup
- VMBT checkpoint clearing when chain is invalid (stale checkpoint would cause KubeVirt to attempt invalid incremental)
- ForceFullBackup defense-in-depth: VMB is created with `ForceFullBackup=true` when chain is broken
- BSL validation runs once per DataUpload (tracked via annotation to avoid redundant S3 queries)
- Graceful degradation: BSL lookup failures are non-fatal, controller falls back to full backup
- Force-full-backup annotation support for user-initiated full backups

Fixes: https://github.com/openshift/oadp-operator/issues/2064
Parent issue: https://github.com/openshift/oadp-operator/issues/2063

## Checkpoint Validation Flow

```
DataUpload (Accepted phase)
    │
    ├── Check force-full-backup annotation → if set, clear VMBT checkpoint, force full
    │
    ├── Check bsl-validated annotation → if set, skip (already validated this DU)
    │
    └── Query BSL for checkpoint chain
        │
        ├── BSL unreachable → leave VMBT as-is, retry next reconcile
        │
        ├── No index.json → clear VMBT checkpoint, force full backup
        │
        ├── Broken chain (missing files) → clear VMBT checkpoint, force full backup
        │
        └── Valid chain found → update VMBT with latest checkpoint (incremental)
```

## Files Changed

| File | Description |
|------|-------------|
| `pkg/uploader/bsl_lookup.go` | BSL checkpoint chain lookup and validation |
| `pkg/uploader/bsl_lookup_test.go` | Tests for checkpoint lookup (729 lines) |
| `pkg/uploader/mock_objectstore.go` | Thread-safe in-memory ObjectStore mock (renamed from test-only) |
| `pkg/common/constants.go` | New annotations: `bsl-validated`, `force-full-backup` |
| `internal/controller/kubevirt_dataupload_controller.go` | BSL validation integration, VMBT checkpoint management, credential handling |
| `internal/controller/kubevirt_dataupload_controller_test.go` | Controller tests (2319 lines added) |
| `config/rbac/role.yaml` | RBAC: added patch/update for VMBT status |
| `config/manager/manager.yaml` | Added /tmp emptyDir volume for credential temp files |

## Manual Testing

Performed on OpenShift cluster (api.mig-shubh-cnv.rhos-psi.cnv-qe.rhood.us:6443) with KubeVirt CBT, ODF/Ceph storage, VM `cirros-test-cont-1` in namespace `cirros-test-cont`, BSL `dpa-test-1`.

All tests used the same DataUpload spec (only `name` changed per test):

```yaml
apiVersion: velero.io/v2alpha1
kind: DataUpload
metadata:
  annotations:
    kubevirt-datamover.io/vm-name: cirros-test-cont-1
    kubevirt-datamover.io/vm-namespace: cirros-test-cont
  name: test-chain-<N>
  namespace: openshift-adp
spec:
  backupStorageLocation: dpa-test-1
  csiSnapshot:
    snapshotClass: ''
    storageClass: ocs-storagecluster-ceph-rbd
    volumeSnapshot: placeholder-not-used
  datamover: kubevirt
  operationTimeout: 10m0s
  snapshotType: CSI
  sourceNamespace: cirros-test-cont
  sourcePVC: cirros-test-cont-1-dv
```

### Test 1: Full Backup (test-chain-1)

**Scenario:** No prior checkpoint chain exists (clean VMBT).
**Result:** PASS - VMB completed with `type: Full`, checkpoint `vmb-test-chain-1-2026-02-16_12-36-16` created.
Phase progression: New -> Accepted -> Prepared -> InProgress -> Completed.

### Test 2: Incremental Backup (test-chain-2)

**Scenario:** Valid checkpoint chain from test-chain-1 exists in BSL.
**Result:** PASS - Checkpoint lookup found valid chain (`latest=vmb-test-chain-1-2026-02-16_12-36-16`), VMB completed with `type: Incremental`.

### Test 3: Full Backup After Broken Chain (test-chain-3)

**Scenario:** Full backup checkpoint folder `vmb-test-chain-1-2026-02-16_12-36-16` manually deleted from S3. Incremental checkpoint left in place.
**Result:** PASS - Controller detected broken chain (`"full backup checkpoint has missing files"`), cleared stale VMBT checkpoint, created VMB with `ForceFullBackup=true`, completed as `type: Full`.

### Test 4: Incremental Backup After Recovery (test-chain-4)

**Scenario:** Chain recovered with new full backup from test-chain-3.
**Result:** PASS - Checkpoint lookup found valid chain (`latest=vmb-test-chain-3-2026-02-16_12-41-30`), VMB completed with `type: Incremental`.

### Test 5: Full Backup After Missing Index (test-chain-5)

**Scenario:** `index.json` manually deleted from S3. All checkpoint data folders left in place.
**Result:** PASS - Controller detected missing index (`"no checkpoint index found (first backup for this VM)"`), cleared stale VMBT checkpoint, created VMB with `ForceFullBackup=true`, completed as `type: Full`.

### Test 6: Incremental Backup After Index Recovery (test-chain-6)

**Scenario:** Chain recovered with new full backup from test-chain-5.
**Result:** PASS - Checkpoint lookup found valid chain (`latest=vmb-test-chain-5-2026-02-16_12-46-05`), VMB completed with `type: Incremental`.

### Test index.json files
The followig file is from the Tests 5-6 (downloaded after Test 6 was ran, because before Test 5 the previous one got deleted):
[index_after_removal.json](https://github.com/user-attachments/files/25340566/index_after_removal.json)

The followig file is from the Tests 1-4 (downloaded after Test 4 was ran):
[index_before_removal.json](https://github.com/user-attachments/files/25340567/index_before_removal.json)

### Summary

| Test | Name | Scenario | Expected | Actual | Result |
|------|------|----------|----------|--------|--------|
| 1 | test-chain-1 | No prior chain | Full | Full | PASS |
| 2 | test-chain-2 | Valid chain exists | Incremental | Incremental | PASS |
| 3 | test-chain-3 | Full backup data deleted from S3 | Full | Full | PASS |
| 4 | test-chain-4 | Chain recovered after Test 3 | Incremental | Incremental | PASS |
| 5 | test-chain-5 | index.json deleted from S3 | Full | Full | PASS |
| 6 | test-chain-6 | Chain recovered after Test 5 | Incremental | Incremental | PASS |

## Test Plan

- [x] Full backup when no checkpoint index exists
- [x] Incremental backup when valid checkpoint chain exists in BSL
- [x] Full backup fallback when checkpoint files are missing from S3
- [x] Full backup fallback when index.json is deleted from S3
- [x] Incremental resumes after chain recovery with new full backup
- [x] BSL validation annotation prevents redundant S3 queries
- [x] VMBT checkpoint cleared when chain is broken
- [x] ForceFullBackup set on VMB as defense-in-depth
- [x] Graceful fallback when BSL credentials are unavailable
- [x] Unit tests for checkpoint lookup, chain validation, controller integration

## Summary by CodeRabbit

* **New Features**
  * Incremental backup support with checkpoint lookup/validation and a new annotation to force full backups; new annotation indicates BSL-validated state.
  * Ability to supply object-store credentials in-memory for backups (no file required).

* **Improvements**
  * Safer incremental fallback to full backups when checkpoint chains are invalid.
  * Expanded RBAC to permit status/checkpoint operations.
  * Datamover pods now mount a tmp emptyDir for reliable temporary file handling.

* **Tests**
  * Extensive new tests covering checkpoint lookup, credential parsing, and backup lifecycle flows.

### DIFF
--- a/config/rbac/datamover_role.yaml
+++ b/config/rbac/datamover_role.yaml
@@ -1,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: datamover-kubevirt-reader
+rules:
+- apiGroups: ["backup.kubevirt.io"]
+  resources: ["virtualmachinebackups", "virtualmachinebackuptrackers"]
+  verbs: ["get", "list", "delete"]
--- a/internal/controller/kubevirt_dataupload_controller.go
+++ b/internal/controller/kubevirt_dataupload_controller.go
@@ -254,40 +254,78 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
 	}
 	logger.Info("Temporary PVC ready", "pvc", pvc.Name)
 
-	// Step 2: Prepare VirtualMachineBackupTracker (recreate from S3 state)
-	vmbt, err := r.prepareVMBackupTracker(ctx, logger, du, vmRef.Name, vmRef.Namespace)
-	if err != nil {
-		logger.Error(err, "Failed to prepare VirtualMachineBackupTracker")
-		return ctrl.Result{}, err
+	// Check if VMB already exists. If it does, the VMBT is already in the right
+	// state (set before VMB creation) and we can skip straight to monitoring.
+	// This avoids deleting/recreating the VMBT while an active VMB references it.
+	vmbName := fmt.Sprintf("%s%s", common.VMBackupNamePrefix, du.Name)
+	existingVMB := &kubevirtbackupv1alpha1.VirtualMachineBackup{}
+	vmbExists := false
+	if err := r.Get(ctx, types.NamespacedName{Name: vmbName, Namespace: vmRef.Namespace}, existingVMB); err == nil {
+		vmbExists = true
+	} else if !errors.IsNotFound(err) {
+		return ctrl.Result{}, fmt.Errorf("failed to check for existing VMB: %w", err)
 	}
-	logger.Info("VirtualMachineBackupTracker ready", "vmbt", vmbt.Name)
 
-	// Step 3: Determine backup mode (full vs incremental).
-	forceFullBackup := r.resolveBackupMode(ctx, logger, du, vmRef)
-
-	// Step 4: Create VirtualMachineBackup if it doesn't exist
-	vmb, created, err := r.ensureVMBackup(ctx, logger, du, vmbt, pvc.Name, vmRef.Namespace, forceFullBackup)
-	if err != nil {
-		// Check if the error is due to another VMB being in progress for the same VM.
-		// KubeVirt's admission webhook only allows one active (non-terminal) VMB per VM.
-		// Requeue with a longer delay instead of returning an error (which causes
-		// exponential backoff retry storm).
-		if strings.Contains(err.Error(), "in progress for source") {
-			logger.Info("Another VirtualMachineBackup is in progress for this VM, will retry",
-				"reason", err.Error())
-			return ctrl.Result{RequeueAfter: RequeueAfterLong}, nil
+	var vmb *kubevirtbackupv1alpha1.VirtualMachineBackup
+	if vmbExists {
+		// VMB already exists — skip VMBT preparation and backup mode resolution.
+		// The VMBT was set up before this VMB was created and should not be disturbed.
+		vmb = existingVMB
+		logger.V(1).Info("VirtualMachineBackup already exists, skipping VMBT preparation", "vmb", vmb.Name)
+	} else {
+		// Step 2: Prepare VirtualMachineBackupTracker (recreate from S3 state)
+		vmbt, err := r.prepareVMBackupTracker(ctx, logger, du, vmRef.Name, vmRef.Namespace)
+		if err != nil {
+			logger.Error(err, "Failed to prepare VirtualMachineBackupTracker")
+			return ctrl.Result{}, err
 		}
-		logger.Error(err, "Failed to ensure VirtualMachineBackup")
-		return ctrl.Result{}, err
+		logger.Info("VirtualMachineBackupTracker ready", "vmbt", vmbt.Name)
+
+		// Step 3: Determine backup mode (full vs incremental).
+		forceFullBackup, checkpointLookup := r.resolveBackupMode(ctx, logger, du, vmRef)
+
+		// Step 3b: Cross-check VMBT checkpoint against BSL chain validation result.
+		// The VMBT's LatestCheckpoint (restored from archived vmbt.json) must agree
+		// with the BSL chain validation (which independently walks the S3 index).
+		// If they diverge, the archived VMBT is stale — force a full backup.
+		if !forceFullBackup && checkpointLookup != nil && checkpointLookup.Found {
+			vmbtCheckpoint := ""
+			if vmbt.Status != nil && vmbt.Status.LatestCheckpoint != nil {
+				vmbtCheckpoint = vmbt.Status.LatestCheckpoint.Name
+			}
+			if vmbtCheckpoint != checkpointLookup.LatestCheckpoint {
+				logger.Info("VMBT checkpoint does not match BSL chain validation result, forcing full backup",
+					"vmbtCheckpoint", vmbtCheckpoint,
+					"bslLatestCheckpoint", checkpointLookup.LatestCheckpoint)
+				forceFullBackup = true
+			}
+		}
+
+		// Step 4: Create VirtualMachineBackup
+		var created bool
+		vmb, created, err = r.ensureVMBackup(ctx, logger, du, vmbt, pvc.Name, vmRef.Namespace, forceFullBackup)
+		if err != nil {
+			// Check if the error is due to another VMB being in progress for the same VM.
+			// KubeVirt's admission webhook only allows one active (non-terminal) VMB per VM.
+			// Requeue with a longer delay instead of returning an error (which causes
+			// exponential backoff retry storm).
+			if strings.Contains(err.Error(), "in progress for source") {
+				logger.Info("Another VirtualMachineBackup is in progress for this VM, will retry",
+					"reason", err.Error())
+				return ctrl.Result{RequeueAfter: RequeueAfterLong}, nil
+			}
+			logger.Error(err, "Failed to ensure VirtualMachineBackup")
+			return ctrl.Result{}, err
+		}
+
+		if created {
+			logger.Info("Created VirtualMachineBackup", "vmb", vmb.Name)
+			// Requeue to check status
+			return ctrl.Result{RequeueAfter: RequeueAfterShort}, nil
+		}
 	}
 
-	if created {
-		logger.Info("Created VirtualMachineBackup", "vmb", vmb.Name)
-		// Requeue to check status
-		return ctrl.Result{RequeueAfter: RequeueAfterShort}, nil
-	}
-
-	// Step 6: Check VMB status
+	// Step 5: Check VMB status
 	if vmb.Status == nil {
 		logger.Info("VirtualMachineBackup status not yet available, requeuing")
 		return ctrl.Result{RequeueAfter: RequeueAfterShort}, nil
@@ -325,15 +363,16 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
 }
 
 // resolveBackupMode determines whether to force a full backup or allow incremental.
-// Returns true when ForceFullBackup should be set on the VMB.
+// Returns (forceFullBackup, checkpointLookup) where checkpointLookup is the BSL
+// chain validation result (nil if BSL was unreachable or validation was skipped).
 // This covers two scenarios:
 //  1. User explicitly requested force-full-backup via annotation on DataUpload.
 //  2. BSL checkpoint validation found a broken chain, requiring a forced full backup.
-func (r *KubeVirtDataUploadReconciler) resolveBackupMode(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmRef *common.VMReference) bool {
+func (r *KubeVirtDataUploadReconciler) resolveBackupMode(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmRef *common.VMReference) (bool, *uploader.CheckpointLookupResult) {
 	// Check if force full backup is requested via annotation.
 	if du.Annotations[common.AnnotationForceFullBackup] == bslValidatedValue {
 		logger.Info("Force full backup requested via annotation, skipping BSL checkpoint lookup")
-		return true
+		return true, nil
 	}
 
 	// Validate checkpoint chain in BSL for incremental backup support.
@@ -342,7 +381,7 @@ func (r *KubeVirtDataUploadReconciler) resolveBackupMode(ctx context.Context, lo
 	// each new backup.
 	if du.Annotations[common.AnnotationBSLValidated] == bslValidatedValue {
 		logger.V(1).Info("BSL validation already completed for this DataUpload")
-		return false
+		return false, nil
 	}
 
 	return r.validateBSLCheckpoint(ctx, logger, du, vmRef)
@@ -350,9 +389,11 @@ func (r *KubeVirtDataUploadReconciler) resolveBackupMode(ctx context.Context, lo
 
 // validateBSLCheckpoint queries the BSL for a valid checkpoint chain and determines
 // whether a forced full backup is required because the chain is broken.
+// Returns (forceFullBackup, checkpointLookup) where checkpointLookup is the BSL
+// chain validation result (nil if BSL was unreachable or lookup failed).
 // It does NOT modify VMBT status — ForceFullBackup on the VMB is the sole mechanism
 // for forcing full backups.
-func (r *KubeVirtDataUploadReconciler) validateBSLCheckpoint(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmRef *common.VMReference) bool {
+func (r *KubeVirtDataUploadReconciler) validateBSLCheckpoint(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmRef *common.VMReference) (bool, *uploader.CheckpointLookupResult) {
 	forceFullBackup := false
 
 	var checkpointLookup *uploader.CheckpointLookupResult
@@ -417,7 +458,7 @@ func (r *KubeVirtDataUploadReconciler) validateBSLCheckpoint(ctx context.Context
 		}
 	}
 
-	return forceFullBackup
+	return forceFullBackup, checkpointLookup
 }
 
 // handlePrepared processes DataUploads in Prepared phase
--- a/internal/controller/kubevirt_dataupload_controller.go
+++ b/internal/controller/kubevirt_dataupload_controller.go
@@ -19,7 +19,9 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"io"
 	"strings"
 	"time"
 
@@ -219,23 +221,28 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
 		return ctrl.Result{}, nil
 	}
 
-	// Step 0: Verify BSL exists before creating any resources.
-	// If BSL is unavailable, creating PVC/VMBT/VMB would be wasteful since
-	// the DataUpload will fail in Prepared phase anyway.
+	// Step 0: Verify BSL exists and is Available before creating any resources.
+	// BSL must be in Available phase — if it's not, fail immediately rather than
+	// creating resources (PVC, VMBT, VMB) that will be wasted.
 	if du.Spec.BackupStorageLocation != "" {
-		if _, err := r.getBackupStorageLocation(ctx, du); err != nil {
-			if errors.IsNotFound(err) {
-				logger.Error(err, "BackupStorageLocation not found")
-				if updateErr := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhaseFailed,
-					fmt.Sprintf("BackupStorageLocation not found: %v", err)); updateErr != nil {
-					return ctrl.Result{}, updateErr
-				}
-				return ctrl.Result{}, nil
+		bslObj, err := r.getBackupStorageLocation(ctx, du)
+		if err != nil {
+			logger.Error(err, "BackupStorageLocation not accessible")
+			if updateErr := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhaseFailed,
+				fmt.Sprintf("BackupStorageLocation not accessible: %v", err)); updateErr != nil {
+				return ctrl.Result{}, updateErr
 			}
-			// Transient error (network, API server issue) - requeue without creating resources
-			logger.Info("BSL temporarily unavailable, will retry before creating resources",
-				"reason", err.Error())
-			return ctrl.Result{RequeueAfter: RequeueAfterLong}, nil
+			return ctrl.Result{}, nil
+		}
+		if bslObj.Status.Phase != velerov1.BackupStorageLocationPhaseAvailable {
+			logger.Error(nil, "BackupStorageLocation is not in Available phase",
+				"bsl", bslObj.Name, "phase", bslObj.Status.Phase)
+			if updateErr := r.updatePhase(ctx, du, velerov2alpha1.DataUploadPhaseFailed,
+				fmt.Sprintf("BackupStorageLocation %q is not available (phase: %s)",
+					bslObj.Name, bslObj.Status.Phase)); updateErr != nil {
+				return ctrl.Result{}, updateErr
+			}
+			return ctrl.Result{}, nil
 		}
 	}
 
@@ -247,16 +254,16 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
 	}
 	logger.Info("Temporary PVC ready", "pvc", pvc.Name)
 
-	// Step 2: Create or get VirtualMachineBackupTracker
-	vmbt, err := r.ensureVMBackupTracker(ctx, logger, du, vmRef.Name, vmRef.Namespace)
+	// Step 2: Prepare VirtualMachineBackupTracker (recreate from S3 state)
+	vmbt, err := r.prepareVMBackupTracker(ctx, logger, du, vmRef.Name, vmRef.Namespace)
 	if err != nil {
-		logger.Error(err, "Failed to ensure VirtualMachineBackupTracker")
+		logger.Error(err, "Failed to prepare VirtualMachineBackupTracker")
 		return ctrl.Result{}, err
 	}
 	logger.Info("VirtualMachineBackupTracker ready", "vmbt", vmbt.Name)
 
-	// Step 3: Determine backup mode (full vs incremental) and update VMBT accordingly.
-	forceFullBackup := r.resolveBackupMode(ctx, logger, du, vmbt, vmRef)
+	// Step 3: Determine backup mode (full vs incremental).
+	forceFullBackup := r.resolveBackupMode(ctx, logger, du, vmRef)
 
 	// Step 4: Create VirtualMachineBackup if it doesn't exist
 	vmb, created, err := r.ensureVMBackup(ctx, logger, du, vmbt, pvc.Name, vmRef.Namespace, forceFullBackup)
@@ -321,22 +328,11 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
 // Returns true when ForceFullBackup should be set on the VMB.
 // This covers two scenarios:
 //  1. User explicitly requested force-full-backup via annotation on DataUpload.
-//  2. BSL checkpoint validation found a broken chain (stale VMBT checkpoint with
-//     no valid S3 data), requiring a forced full backup as defense-in-depth.
-func (r *KubeVirtDataUploadReconciler) resolveBackupMode(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmbt *kubevirtbackupv1alpha1.VirtualMachineBackupTracker, vmRef *common.VMReference) bool {
+//  2. BSL checkpoint validation found a broken chain, requiring a forced full backup.
+func (r *KubeVirtDataUploadReconciler) resolveBackupMode(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmRef *common.VMReference) bool {
 	// Check if force full backup is requested via annotation.
 	if du.Annotations[common.AnnotationForceFullBackup] == bslValidatedValue {
 		logger.Info("Force full backup requested via annotation, skipping BSL checkpoint lookup")
-
-		vmbtHasCheckpoint := vmbt.Status != nil && vmbt.Status.LatestCheckpoint != nil &&
-			vmbt.Status.LatestCheckpoint.Name != ""
-		if vmbtHasCheckpoint {
-			if err := r.clearVMBTCheckpoint(ctx, logger, vmbt); err != nil {
-				logger.Error(err, "Failed to clear VMBT checkpoint for forced full backup")
-			} else {
-				logger.Info("Cleared VMBT checkpoint for forced full backup")
-			}
-		}
 		return true
 	}
 
@@ -349,29 +345,29 @@ func (r *KubeVirtDataUploadReconciler) resolveBackupMode(ctx context.Context, lo
 		return false
 	}
 
-	return r.validateBSLCheckpoint(ctx, logger, du, vmbt, vmRef)
+	return r.validateBSLCheckpoint(ctx, logger, du, vmRef)
 }
 
-// validateBSLCheckpoint queries the BSL for a valid checkpoint chain and updates
-// the VMBT accordingly. Returns true when a forced full backup is required because
-// the checkpoint chain was found to be broken.
-func (r *KubeVirtDataUploadReconciler) validateBSLCheckpoint(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmbt *kubevirtbackupv1alpha1.VirtualMachineBackupTracker, vmRef *common.VMReference) bool {
+// validateBSLCheckpoint queries the BSL for a valid checkpoint chain and determines
+// whether a forced full backup is required because the chain is broken.
+// It does NOT modify VMBT status — ForceFullBackup on the VMB is the sole mechanism
+// for forcing full backups.
+func (r *KubeVirtDataUploadReconciler) validateBSLCheckpoint(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmRef *common.VMReference) bool {
 	forceFullBackup := false
 
 	var checkpointLookup *uploader.CheckpointLookupResult
 	bsl, bslErr := r.getBackupStorageLocation(ctx, du)
 	if bslErr != nil {
-		// BSL lookup failure is non-fatal. VMBT is left as-is: if it already has
-		// a valid checkpoint, KubeVirt will use it for incremental backup.
-		// Validation will be retried on the next reconcile.
+		// BSL lookup failure is non-fatal. Validation will be retried on the
+		// next reconcile.
 		logger.Info("BSL not available for checkpoint lookup, skipping validation",
 			"reason", bslErr.Error())
 	} else {
 		var err error
 		checkpointLookup, err = r.lookupCheckpointFromBSL(ctx, bsl, vmRef.Namespace, vmRef.Name)
 		if err != nil {
-			// Checkpoint lookup failure is non-fatal. VMBT is left as-is
-			// and validation will be retried on the next reconcile.
+			// Checkpoint lookup failure is non-fatal. Validation will be retried
+			// on the next reconcile.
 			logger.Info("Checkpoint lookup failed, skipping validation",
 				"reason", err.Error())
 		} else if checkpointLookup != nil {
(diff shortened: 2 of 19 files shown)


internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -222,9 +236,21 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
-	// Step 3: Create VirtualMachineBackup if it doesn't exist
-	vmb, created, err := r.ensureVMBackup(ctx, logger, du, vmbt, pvc.Name, vmRef.Namespace)
+	// Step 3: Determine backup mode (full vs incremental) and update VMBT accordingly.
+	forceFullBackup := r.resolveBackupMode(ctx, logger, du, vmbt, vmRef)
+
+	// Step 4: Create VirtualMachineBackup if it doesn't exist
+	vmb, created, err := r.ensureVMBackup(ctx, logger, du, vmbt, pvc.Name, vmRef.Namespace, forceFullBackup)
 	if err != nil {
+		// Check if the error is due to another VMB being in progress for the same VM.
+		// KubeVirt's admission webhook only allows one VMB per VM at a time.
```
sseago: Is this confirmed? Also, the two lines above seem to imply different things. 245 suggests "allows one non-terminal VMP per VM at the same time", but 246 suggests that even completed VMBs would prevent future VMB creation.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -272,6 +298,127 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
 	return ctrl.Result{RequeueAfter: RequeueAfterShort}, nil
 }
 
+// resolveBackupMode determines whether to force a full backup or allow incremental.
+// Returns true when ForceFullBackup should be set on the VMB.
+// This covers two scenarios:
+//  1. User explicitly requested force-full-backup via annotation on DataUpload.
```
sseago, line 369: Shouldn't the annotation be on the backup CR? User doesn't create DataUpload so the user won't be putting any annotations there.

pkg/common/constants.go
```diff
@@ -31,6 +31,18 @@ const (
...
+	// performed for this DataUpload. This prevents redundant S3 queries on
+	// every reconcile loop iteration while still ensuring validation runs
+	// once per DataUpload.
+	AnnotationBSLValidated = "kubevirt-datamover.io/bsl-validated"
+
+	// AnnotationForceFullBackup, when set to "true" on a DataUpload, forces
+	// a full backup even when a valid incremental checkpoint chain exists.
+	// The VMBT checkpoint is cleared and the VMB is created with
+	// ForceFullBackup=true. The new checkpoint replaces the old one in BSL.
+	AnnotationForceFullBackup = "kubevirt-datamover.io/force-full-backup"
```
shubham-pampattiwar, line 45: The controller checks for `AnnotationForceFullBackup` on DataUpload CRs to force full backups, but in production users create Velero Backup CRs and not DataUpload CRs - the kubevirt-datamover-plugin creates the DataUploads automatically.

How are users supposed to trigger a forced full backup? Do they:
1. Annotate the Backup CR and the plugin propagates it to DataUpload?
2. Manually patch the DataUpload after the plugin creates it?

I believe we need some plugin work to make this functional ? Lets add a TODO and create an issue for this if thats the case.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -272,6 +298,127 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
+
+// validateBSLCheckpoint queries the BSL for a valid checkpoint chain and updates
+// the VMBT accordingly. Returns true when a forced full backup is required because
+// the checkpoint chain was found to be broken.
+func (r *KubeVirtDataUploadReconciler) validateBSLCheckpoint(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmbt *kubevirtbackupv1alpha1.VirtualMachineBackupTracker, vmRef *common.VMReference) bool {
+	forceFullBackup := false
+
+	var checkpointLookup *uploader.CheckpointLookupResult
+	bsl, bslErr := r.getBackupStorageLocation(ctx, du)
+	if bslErr != nil {
```
shubham-pampattiwar, line 401: `validateBSLCheckpoint` treats BSL unavailability as non-fatal and continues, but `handlePrepared` fails the DataUpload for the same error.

This means if BSL is unavailable during Accepted phase, the controller will:
1. Gracefully continue and create VMBT/VMB/PVC
2. Wait for VMB to complete
3. Transition to Prepared
4. Then fail when trying to access BSL again

Why not fail fast in Accepted phase instead of doing unnecessary work?

Additionally, Velero validates BSL availability before uploading K8s objects. Following the same pattern, we should fail the DataUpload early if BSL isn't available for the datamover upload.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -896,3 +1022,223 @@ func extractPodFailureMessage(pod *corev1.Pod) string {
...
+		CredentialName: credName,
+		CredentialKey:  credKey,
+	}, nil
+}
+
+// lookupCheckpointFromBSL reads the VM's checkpoint index from the BSL and returns
+// the latest valid checkpoint for incremental backup support.
+// This method initializes an object store client, reads the BSL credentials from
+// the secret, and queries the checkpoint index.
+func (r *KubeVirtDataUploadReconciler) lookupCheckpointFromBSL(ctx context.Context, bsl *velerov1.BackupStorageLocation, vmNamespace, vmName string) (*uploader.CheckpointLookupResult, error) {
```
shubham-pampattiwar, line 1255: This func creates temporary files for BSL credentials. Consider adding a `Credentials []byte` field to `UploaderConfig` alongside the existing `CredentialsFile string`. This Eliminates temp file I/O and cleanup and Better security (credentials stay in memory). Thoughts ?

pkg/uploader/bsl_lookup.go
```diff
@@ -0,0 +1,225 @@
...
+
+		// Try the previous checkpoint in the chain
+		currentTargetID = chain[brokenAt-1].ID
+	}
+}
+
+// validateCheckpointFiles checks that all qcow2 files in a checkpoint exist in BSL.
+// The ObjectPath in each CheckpointFile is expected to be a relative key (without
+// the store prefix), matching how the uploader writes them via the same ObjectStore.
+func validateCheckpointFiles(store velero.ObjectStore, bucket string, files []CheckpointFile) error {
```
shubham-pampattiwar: This func checks files sequentially. For VMs with multiple disks, this can be slow:
```go
// Current: Sequential (3 disks × 100ms = 300ms)
for _, f := range files {
    exists, err := store.ObjectExists(bucket, f.ObjectPath)
}
```
Please consider parallelizing this: eg
```go
g, ctx := errgroup.WithContext(ctx)
for _, f := range files {
    file := f
    g.Go(func() error {
        exists, err := store.ObjectExists(bucket, file.ObjectPath)
        // ... error handling
    })
}
return g.Wait()

```
This would improve performance for VMs with long checkpoint chains and multiple disks. WDYT ?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -896,3 +1022,223 @@ func extractPodFailureMessage(pod *corev1.Pod) string {
...
+	}
+
+	logger.Info("Cleared VMBT checkpoint", "vmbt", vmbt.Name)
+	return nil
+}
+
+// updateVMBTCheckpoint updates the VirtualMachineBackupTracker status with the latest
+// checkpoint from BSL. This enables KubeVirt to perform an incremental backup
+// based on the given checkpoint.
+func (r *KubeVirtDataUploadReconciler) updateVMBTCheckpoint(ctx context.Context, logger logr.Logger, vmbt *kubevirtbackupv1alpha1.VirtualMachineBackupTracker, checkpointName string) error {
```
shubham-pampattiwar: This controller creates and modifies VMBT.Status (updateVMBTCheckpoint, clearVMBTCheckpoint). However:
- VMBT/VMB are KubeVirt-owned CRs (kubevirt.io/api/backup/v1alpha1)
- Issue #12 plans to delete VMBT/VMB after backup and store in S3

Should this controller modify VMBT.Status, or only read it? I hink we will have to treat BSL (S3) as source of truth for checkpoint state rather than maintaining state in VMBT.

internal/controller/kubevirt_dataupload_controller.go
mpryc: I was getting the following error from the admission webhook when attempting to create new DU. The webhook was saying another VMB already exists, when checked it was never finished so that's why above check: 

```shell
ERROR	Failed to ensure VirtualMachineBackup	{"controller": "kubevirt-dataupload", "controllerGroup": "velero.io", "controllerKind": "DataUpload", "DataUpload": {"name":"datamover-mpryc-du-newest-second-finesthree","namespace":"openshift-adp"}, "namespace": "openshift-adp", "name": "datamover-mpryc-du-newest-second-finesthree", "reconcileID": "2fa614d5-379c-4e73-abbb-b9d9f9663357", "error": "failed to create VirtualMachineBackup: admission webhook \"virtualmachinebackup-validator.backup.kubevirt.io\" denied the request: VirtualMachineBackup \"vmb-datamover-mpryc-du-newest-second-finesthree\" already exists"}
github.com/migtools/kubevirt-datamover-controller/internal/controller.(*KubeVirtDataUploadReconciler).handleAccepted
	/workspace/internal/controller/kubevirt_dataupload_controller.go:333
github.com/migtools/kubevirt-datamover-controller/internal/controller.(*KubeVirtDataUploadReconciler).Reconcile
	/workspace/internal/controller/kubevirt_dataupload_controller.go:134
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.0/pkg/internal/controller/controller.go:222
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.0/pkg/internal/controller/controller.go:479
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.0/pkg/internal/controller/controller.go:438
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func1.1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.23.0/pkg/internal/controller/controller.go:313
```

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -272,6 +298,127 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
+	if checkpointLookup != nil {
+		if du.Annotations == nil {
+			du.Annotations = make(map[string]string)
+		}
+		du.Annotations[common.AnnotationBSLValidated] = bslValidatedValue
+		if err := r.Update(ctx, du); err != nil {
+			// Non-fatal: worst case we re-run BSL validation on next reconcile
+			logger.Info("Failed to set BSL validated annotation, will retry",
+				"reason", err.Error())
+		}
```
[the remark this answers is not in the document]
mpryc: No change needed.

The `r.Update(ctx, du)` at line 432 operates on the same `du` object that was fetched at the top of `Reconcile()` (line 110-114). Within a single reconcile call, the `du` object has the correct `resourceVersion` because no other code path updates the DataUpload object before this point in the `handleAccepted` flow.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -896,3 +1022,223 @@ func extractPodFailureMessage(pod *corev1.Pod) string {
...
+	}
+	vmbt.Status.LatestCheckpoint = &kubevirtbackupv1alpha1.BackupCheckpoint{
+		Name:         checkpointName,
+		CreationTime: &now,
+	}
+
+	if err := r.Status().Update(ctx, vmbt); err != nil {
+		return fmt.Errorf("failed to update VMBT %s/%s status with checkpoint %s: %w",
+			vmbt.Namespace, vmbt.Name, checkpointName, err)
+	}
```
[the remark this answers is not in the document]
mpryc: Same reasoning as previous comment. The `vmbt` object is fetched via `ensureVMBackupTracker()` earlier in `handleAccepted()` (line 252). Between that fetch and the status update in `clearVMBTCheckpoint`/`updateVMBTCheckpoint`, no other code path in this controller updates the VMBT.

pkg/uploader/bsl_lookup.go
```diff
@@ -0,0 +1,225 @@
...
+			continue
+		}
+
+		exists, err := store.ObjectExists(bucket, f.ObjectPath)
+		if err != nil {
+			return fmt.Errorf("failed to check file %s: %w", f.ObjectPath, err)
+		}
+		if !exists {
+			return fmt.Errorf("file %s not found in BSL", f.ObjectPath)
+		}
```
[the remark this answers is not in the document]
mpryc: Agree on that. Implemented in the last commit. An empty `ObjectPath` cannot legitimately occur. The backup flow guarantees a qcow2 file per CBT-eligible disk. I had doubts around backups where no qcow2 changed, but (see point 2):

1. **KubeVirt generates the libvirt backup XML** in `generateDomainBackup()` ([`pkg/virt-launcher/virtwrap/storage/backup.go:204-256`](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-launcher/virtwrap/storage/backup.go#L204)). It iterates all domain disks (line 216) and for each CBT-eligible disk (`disk.Source.DataStore != nil`, line 227), sets `backupDisk.Backup = "yes"` (line 228) with a target qcow2 file path via `targetQCOW2File()` (line 232, defined at line 264).
2. **Libvirt creates a qcow2 for every disk** marked `Backup = "yes"` in the backup XML. Even when there are zero changed blocks (incremental with no writes), libvirt produces a qcow2 file (it will simply have no data extents). Disks not eligible for CBT are marked `Backup = "no"` (line 241) and are excluded entirely.

Since every CBT-eligible disk always produces a qcow2 file, and the uploader always sets `ObjectPath` for every file it uploads, an empty `ObjectPath` in a `CheckpointFile` can only indicate index corruption (e.g., manual editing, partial write, JSON marshaling bug).

internal/controller/kubevirt_dataupload_controller.go
mpryc: The comment is **NOT** accurate.

KubeVirt's admission webhook (`virtualmachinebackup-validator`) rejects creation of a new VMB when there is an existing **non-terminal** VMB for the same VM source.

Completed VMBs do not block new ones - only in-progress ones do. Updated the comment to be precise.

internal/controller/kubevirt_dataupload_controller.go
mpryc: I agree that the DataUpload is and should not be user facing CR, however:

- For **Testing** or debugging purposes I think it's perfectly fine to have such annotation, especially for the cases where we want to create one without plugin enabled. 
- It's also set by the controller when a broken chain is detected, as a safety net alongside VMBT checkpoint clearing.

As @shubham-pampattiwar wrote https://github.com/migtools/kubevirt-datamover-controller/pull/13#discussion_r2814462577 we could include that (but let's not do it as part of the phase4) as a user-facing annotation on the Backup CR in a future phase and propagate it down to the current DataUpload mechanism.

pkg/common/constants.go
mpryc: Yes, I think this is currently not in the design and I think we can keep it as issue to resolve outside of this PR. See my comment on https://github.com/migtools/kubevirt-datamover-controller/pull/13#discussion_r2816809829

internal/controller/kubevirt_dataupload_controller.go
mpryc: Added a "Step 0" BSL availability check at the start of `handleAccepted`, before any resource creation (PVC, VMBT, VMB). When `du.Spec.BackupStorageLocation` is set (always the case in production), the controller verifies the BSL CR exists via a lightweight Kubernetes API GET:

- **BSL not found** (`errors.IsNotFound`): Permanent failure - DataUpload is failed immediately with a clear error message. No PVC/VMBT/VMB are created.
- **Transient error** (API server issue, network): Requeues with `RequeueAfterLong` without creating resources. BSL will be checked again on next reconcile.
- **BSL exists**: Proceeds to Step 1+ as before. The deeper checkpoint chain validation (`validateBSLCheckpoint`) still runs in Step 3 after VMBT is available.

This approach keeps `resolveBackupMode`/`validateBSLCheckpoint` signatures unchanged. The Step 0 check is purely an existence check (does the BSL CR exist?), while Step 3 does the full chain validation (are credentials available? does the checkpoint chain have valid S3 data?).

Test cases added:
- `TestHandleAccepted_BSLNotFound_FailsDataUpload`: BSL specified but doesn't exist, DataUpload fails immediately, no PVCs created
- `TestHandleAccepted_BSLTransientError_Requeues`: BSL not found (simulated via fake client), DataUpload fails, no PVCs created

pkg/uploader/bsl_lookup.go
mpryc: fixed

internal/controller/kubevirt_dataupload_controller.go
mpryc: YES x 3. Changed.

internal/controller/kubevirt_dataupload_controller.go
mpryc: The current implementation IMO is correct - the controller must both read and write VMBT.Status.LatestCheckpoint.

1. KubeVirt's API requires it. VMBT.Status.LatestCheckpoint is the interface KubeVirt exposes to determine the incremental base. isIncrementalBackup() checks this field, and if set, passes it to libvirt as the parent checkpoint. https://github.com/kubevirt/kubevirt/blob/main/pkg/storage/cbt/backup.go#L312-L317
  
2. Clearing stale checkpoints prevents unrestorable backups. If S3 data for a previous checkpoint is deleted, the controller clears LatestCheckpoint to force a full backup. Using only VMB.Spec.ForceFullBackup would be fragile - if the VMB fails and for example BSL becomes temporarily unavailable on retry, the stale checkpoint remains in VMBT. KubeVirt would then produce an incremental backup whose chain is broken in S3, making it unrestorable.

3. Setting checkpoints enables incremental after VMBT recreation. When VMBT is deleted and recreated (planned for Phase 5), the fresh VMBT has no checkpoint. Without writing the validated checkpoint from BSL lookup, every post-recreation backup would be a full backup ??

4. No write conflict. The writes are sequential: our controller writes VMBT before VMB creation (Accepted phase), KubeVirt reads it when processing the VMB, KubeVirt writes back the new checkpoint after completion (updateBackupTracker()). The two controllers never write concurrently.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -896,3 +1043,206 @@ func extractPodFailureMessage(pod *corev1.Pod) string {
...
+	if bucket == "" {
+		return nil, fmt.Errorf("BSL %s has no bucket configured", bsl.Name)
+	}
+
+	// Add our datamover prefix to the BSL prefix
+	if prefix != "" {
+		prefix = prefix + "-" + common.DatamoverBSLPrefix
+	} else {
+		prefix = common.DatamoverBSLPrefix
+	}
```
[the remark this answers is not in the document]
mpryc: The prefix from extractBSLConfig is used as the S3ObjectStore.prefix field. The only place it becomes part of an S3 key is fullKey() at line 139, which already does strings.TrimSuffix(s.prefix, "/") before appending "/". So even if the BSL prefix were "velero/", the constructed prefix would be "velero/-kubevirt-datamover" and fullKey would produce "velero/-kubevirt-datamover/checkpoints/..." - the trailing slash is handled by fullKey, not by the prefix construction.

This is a theoretical concern about input that doesn't occur in practice, and even if it did, the result is a valid (if slightly ugly) S3 key that works correctly. Not worth adding defensive code for.

pkg/common/constants.go
weshayutin: @mpryc @shubham-pampattiwar I suspect forcing a full backup will ALSO come into play in velero PR #9528 and I don't see it addressed there.  The most logical place for the user is a filed in the backup spec.  Perhaps that begins with a label on the velero backup spec and graduates to the crd.

internal/controller/kubevirt_dataupload_controller.go
weshayutin: If the BSL is not "available" backup should immediately fail. The BSL is always exepcted to be available

pkg/common/constants.go
sseago: @weshayutin yes. I believe that's addressed in the design. For 1.6 we're using a forceFullBackup annotation on the backup CR, with the intent of deprecating this annotation once we get a backup spec field which will force full for both kubevirt and velero datamover.

shubham-pampattiwar:
Latest discussions pertaining to phase 4 updates: https://github.com/migtools/kubevirt-datamover-controller/issues/12#issuecomment-3916767936

internal/controller/kubevirt_dataupload_controller.go
sseago: @mpryc The other thing you could do to avoid having to propagate it is in some future commit, check for the force full annotation both here and on the backup CR. Either is fine, as long as putting it on the backup CR results in all VMs forcing full backup.

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -643,20 +817,45 @@ func (r *KubeVirtDataUploadReconciler) ensureTempPVC(ctx context.Context, logger
 	return pvc, nil
 }
 
-// ensureVMBackupTracker creates or retrieves the VirtualMachineBackupTracker for the VM
-func (r *KubeVirtDataUploadReconciler) ensureVMBackupTracker(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmName, vmNamespace string) (*kubevirtbackupv1alpha1.VirtualMachineBackupTracker, error) {
-	// Use a consistent name based on VM name for tracking across backups
+// prepareVMBackupTracker creates a VirtualMachineBackupTracker for the VM,
+// restoring the LatestCheckpoint from the archived VMBT in S3 if available.
+// Any existing VMBT is deleted first to avoid conflicts (S3 is the source of truth).
+func (r *KubeVirtDataUploadReconciler) prepareVMBackupTracker(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmName, vmNamespace string) (*kubevirtbackupv1alpha1.VirtualMachineBackupTracker, error) {
```
shubham-pampattiwar, line 864: I noticed that `prepareVMBackupTracker` unconditionally deletes and recreates the VMBT on every reconcile of the Accepted phase. On the first pass this makes sense, but on subsequent reconciles (e.g., while we're polling VMB status), the VMB already exists and references the VMBT. I'm a bit worried about deleting the VMBT out from under an active VMB, do you know if KubeVirt handles that gracefully?

One thought, since VMBT is always created before VMB, if the VMB for this DataUpload already exists, the VMBT should already be in the right state. Could we skip `prepareVMBackupTracker` in that case and just fetch the existing VMBT instead? Curious what you think.

pkg/uploader/bsl_lookup.go
```diff
@@ -0,0 +1,248 @@
...
+// If files are missing somewhere in the chain, it iteratively tries shorter chains
+// by falling back to earlier checkpoints until a valid chain is found or all
+// options are exhausted.
+func validateCheckpointChain(
+	ctx context.Context, store velero.ObjectStore, bucket string,
+	checkpoints []CheckpointEntry, targetID string,
+) (*CheckpointLookupResult, error) {
+	currentTargetID := targetID
+
+	for {
```
shubham-pampattiwar, line 144: This loop walks backward through the chain when it finds missing files. If the index.json were ever corrupted with a cycle in parent references (e.g., `cp-002` → `cp-003` → `cp-002`), this could loop indefinitely. Unlikely since we control the index writes, but would a simple max iteration guard like `len(checkpoints)` make sense here as a safety net?

config/rbac/datamover_role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
shubham-pampattiwar, line 1: Question on the datamover role/role binding files, since this controller is intended to be installed via the OADP operator, would the operator be responsible for managing RBAC for the velero ServiceAccount? If so, these files might not belong here and could potentially conflict with what the operator sets up. Also noticed the binding uses `namespace: system` which would need to be patched to the actual OADP namespace.

Would it make more sense to track the datamover pod RBAC as a requirement on the OADP operator side instead of shipping it here?

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -214,17 +254,29 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
...
-		logger.Error(err, "Failed to ensure VirtualMachineBackupTracker")
+		logger.Error(err, "Failed to prepare VirtualMachineBackupTracker")
 		return ctrl.Result{}, err
 	}
 	logger.Info("VirtualMachineBackupTracker ready", "vmbt", vmbt.Name)
 
-	// Step 3: Create VirtualMachineBackup if it doesn't exist
-	vmb, created, err := r.ensureVMBackup(ctx, logger, du, vmbt, pvc.Name, vmRef.Namespace)
+	// Step 3: Determine backup mode (full vs incremental).
+	forceFullBackup := r.resolveBackupMode(ctx, logger, du, vmRef)
```
shubham-pampattiwar: As discussed in the design comment (https://github.com/migtools/kubevirt-datamover-controller/issues/12#issuecomment-3916767936), before creating the VMB we should verify the checkpoint on the VMBT status matches what we expect from the S3 chain validation. Currently `prepareVMBackupTracker` restores the VMBT checkpoint from the archived `vmbt.json`, and `validateBSLCheckpoint` independently determines the latest valid checkpoint from the chain, but there's no cross-check between these two values before we proceed to create the VMB.

Could we add a verification step after both lookups to confirm the VMBT's `LatestCheckpoint` matches the chain validation result, and force a full backup if they diverge?

```go
// Verify VMBT checkpoint matches BSL chain validation result.
// vmbtCheckpoint: restored from archived vmbt.json in S3 (set by prepareVMBackupTracker)
// bslCheckpoint: latest valid checkpoint from S3 chain validation (set by validateBSLCheckpoint)
// If they diverge, force a full backup to avoid an inconsistent incremental.
if !forceFullBackup && checkpointLookup != nil && checkpointLookup.Found {
    vmbtCheckpoint := ""
    if vmbt.Status != nil && vmbt.Status.LatestCheckpoint != nil {
        vmbtCheckpoint = vmbt.Status.LatestCheckpoint.Name
    }
    if vmbtCheckpoint != checkpointLookup.LatestCheckpoint {
        logger.Info("VMBT checkpoint does not match BSL chain validation result, forcing full backup",
            "vmbtCheckpoint", vmbtCheckpoint,
            "bslLatestCheckpoint", checkpointLookup.LatestCheckpoint)
        forceFullBackup = true
    }
}
```

Note: `checkpointLookup` is currently scoped inside `validateBSLCheckpoint` and not returned to `handleAccepted`, so this would need `resolveBackupMode` to return the lookup result as well.

pkg/uploader/run.go
```diff
@@ -81,8 +87,21 @@ func Run(ctx context.Context) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create K8s client: %w", err)
+	}
+
+	// Archive VMB/VMBT CRs to S3 (FATAL if fails — index references must be valid)
+	archived, err := archiveKubeResources(ctx, store, k8sClient, config)
+	if err != nil {
+		return fmt.Errorf("failed to archive Kube resources: %w", err)
+	}
+	fmt.Println("Kube resources archived to S3")
```
shubham-pampattiwar: `run.go` uses `fmt.Println`/`fmt.Printf` for logging, worth switching to a structured logger in a follow-up.

mpryc: Agreed. Switched to a structured logger in the last separate commit to see what's changed. 

Captured the datamover pod logs and confirmed they
are now structured JSON with the `"uploader"` logger name:
```shell
{"level":"info","ts":"2026-02-19T13:00:40Z","logger":"uploader","msg":"Starting kubevirt datamover uploader"}
{"level":"info","ts":"2026-02-19T13:00:40Z","logger":"uploader","msg":"Config loaded","vm":"cirros-test-cont/cirros-test-cont-1","checkpoint":"vmb-test-logging-3-2026-02-19_13-00-03","type":"Incremental"}
{"level":"info","ts":"2026-02-19T13:00:40Z","logger":"uploader","msg":"Uploading file","filename":"vmb-test-logging-3-volume0.qcow2","size":196616,"objectPath":"..."}
{"level":"info","ts":"2026-02-19T13:00:41Z","logger":"uploader","msg":"Uploaded qcow2 files","count":1}
{"level":"info","ts":"2026-02-19T13:00:41Z","logger":"uploader","msg":"Archived VMB","path":"..."}
{"level":"info","ts":"2026-02-19T13:00:42Z","logger":"uploader","msg":"Deleted VMB from cluster","vmb":"cirros-test-cont/vmb-test-logging-3"}
{"level":"info","ts":"2026-02-19T13:00:42Z","logger":"uploader","msg":"Upload completed successfully"}
```

internal/controller/kubevirt_dataupload_controller.go
mpryc: Very well said. I've missed that, but actually it's required ! Added a cross-validation step.

Also performed the following manual test to ensure it catches that problem:

## Scenario A: Normal incremental backup (happy path)

1. Created DataUpload `test-scenario-a` targeting cirros-test-cont-1.
2. Confirmed the controller transitioned it to Accepted phase.
3. On the first reconcile of Accepted phase, the controller:
   - Created the temporary PVC
   - Found the archived VMBT in BSL with latestCheckpoint
     `vmb-test-du-newtest-9-2026-02-18_15-35-29`
   - Created a new VMBT with that checkpoint restored
   - Ran BSL chain validation which found a valid 2-checkpoint chain with
     the same latest checkpoint `vmb-test-du-newtest-9-2026-02-18_15-35-29`
   - The VMBT checkpoint and BSL chain checkpoint matched, so no
     cross-check mismatch was triggered
   - Created VMB without ForceFullBackup (incremental allowed)
4. On subsequent reconciles (polling VMB status), the controller logged
   `"VirtualMachineBackup already exists, skipping VMBT preparation"` and
   did NOT log `"Deleting existing VMBT before recreation"`. This confirms
   Comment 1 fix is working: the VMBT is not disturbed while the VMB
   references it.
5. The backup completed successfully as an incremental backup.

## Scenario B: Stale vmbt.json forces full backup (validates problem that is in this comment)

1. After successful scenario completed, the latest checkpoint in index.json was
   `vmb-test-scenario-a-2026-02-19_12-02-30`.
2. Downloaded the vmbt.json for that checkpoint from S3 and tampered with
   it: changed `status.latestCheckpoint.name` from
   `vmb-test-scenario-a-2026-02-19_12-02-30` to
   `vmb-test-du-newtest-7-2026-02-18_15-28-24` (an older checkpoint).
   Uploaded the tampered vmbt.json back to S3.
3. Created DataUpload `test-scenario-b` targeting the same VM.
4. On the first reconcile of Accepted phase, the controller:
   - Found the archived VMBT in BSL with the stale latestCheckpoint
     `vmb-test-du-newtest-7-2026-02-18_15-28-24` (from tampered vmbt.json)
   - Created a new VMBT with that stale checkpoint restored
   - Ran BSL chain validation which found a valid 3-checkpoint chain with
     latest `vmb-test-scenario-a-2026-02-19_12-02-30` (from index.json)
   - The cross-check detected the mismatch and logged:
     `"VMBT checkpoint does not match BSL chain validation result, forcing full backup"`
     with vmbtCheckpoint=vmb-test-du-newtest-7-... vs
     bslLatestCheckpoint=vmb-test-scenario-a-...
   - Created VMB with ForceFullBackup=true
5. On subsequent reconciles, the fix also worked:
   `"VirtualMachineBackup already exists, skipping VMBT preparation"`.
6. The backup completed successfully.
7. Verified in S3 index.json that the new checkpoint
   `vmb-test-scenario-d-2026-02-19_12-06-39` was recorded as Type: Full
   with no parent and file size 157614080 bytes (full disk image), compared
   to ~196616 bytes for incremental backups.

config/rbac/datamover_role.yaml
mpryc: The `namespace: system` in `datamover_role_binding.yaml` is the standard kustomize placeholder right?

We use this in our OADP Non-Admin controller uses the same `namespace: system` in its [role_binding.yaml](https://github.com/migtools/oadp-non-admin/blob/oadp-dev/config/rbac/role_binding.yaml#L19).

The RBAC files are defined here as the source of truth for what permissions the datamover pod needs. The OADP operator's Makefile copies these RBAC manifests from this repo, and the operator's own [kustomization.yaml](https://github.com/openshift/oadp-operator/blob/oadp-dev/config/default/kustomization.yaml#L2) applies the correct namespace when deploying them. This is the same pattern used for the Non-Admin controller's RBAC.

pkg/uploader/bsl_lookup.go
mpryc: The actual infinite loop risk was in `buildCheckpointChain` (`pkg/uploader/run.go`), which follows parent pointers without any cycle detection. Added a `visited` set that breaks the loop if a checkpoint ID is encountered twice. Two unit test cases were added covering a 3-node cycle (`cp-001 -> cp-003 -> cp-002 -> cp-001`) and a self-referencing checkpoint (`cp-001 -> cp-001`). Both pass.

internal/controller/kubevirt_dataupload_controller.go
mpryc: On subsequent reconciles of the Accepted phase (while polling VMB status), the VMBT was being deleted and recreated even though the active VMB references it. Now `handleAccepted` checks if the VMB for this DataUpload already exists before calling `prepareVMBackupTracker`. If the VMB is present, we skip VMBT preparation entirely and go straight to VMB status monitoring.

Verified on cluster: created DataUpload `test-scenario-a`. On the first reconcile the controller created the VMBT and VMB normally. On all subsequent reconciles while the VMB was in progress, the controller logged `"VirtualMachineBackup already exists, skipping VMBT preparation"` and never logged `"Deleting existing VMBT before recreation"`. The backup completed successfully as an incremental.

internal/controller/kubevirt_dataupload_controller.go
mpryc: Done. At line 224-246, Step 0 in handleAccepted checks the BSL phase before creating any resources. If the BSL is not in Available phase, the DataUpload is immediately set to Failed with a message like BackupStorageLocation "dpa-test-1" is not available (phase: Unavailable). No PVC, VMBT, or VMB is created.

internal/controller/kubevirt_dataupload_controller.go
mpryc: I like the idea of the backup cr user-facing annotation, but this won't change how controller is implemented. The controller already checks annotations on the DU.

The only change needed is in the plugin. The DU should set the annotation from the Backup one and we are done with this user facing feature. Let's keep current controller logic as is and the DU  will be internal mechanism.

weshayutin (approved):
/LGTM

sseago (approved):

PR #21, opened 2026-02-26
fix: Update referenceBy field with all backups involved

### LINKED ISSUE #18: Update referencedBy field in checkpoints index.json file with all the corrects backups involved in the checkpoint chain
The checkpoints per-vm index.json file should, for each checkpoint, list all backups that depend on it, not just the first. See the design doc for an example:

https://github.com/openshift/oadp-operator/blob/oadp-dev/docs/design/kubevirt-datamover.md?plain=1#L136

mpryc:
Fixes: #18

Adds logic and updates the referenceBy to include all backups that are involved in the chain.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed circular reference handling in checkpoint chains
  * Ensured backup references are preserved during incremental backups

* **Improvements**
  * Enhanced propagation of backup references across checkpoint hierarchies for improved garbage collection reliability
  * Improved parent-child relationship integrity in checkpoint chains

* **Tests**
  * Added comprehensive test coverage for backup reference handling and chain propagation scenarios

### DIFF
--- a/pkg/uploader/run.go
+++ b/pkg/uploader/run.go
@@ -356,6 +356,14 @@ func updateVMIndex(
 	found := false
 	for i, cp := range vmIndex.Checkpoints {
 		if cp.ID == checkpoint.ID {
+			// Guard against self-referential parent. This can happen when an
+			// incremental backup re-runs: the checkpoint is already the last
+			// entry, so validateCheckpointChain returns its own ID as the
+			// latest valid checkpoint, causing Parent to point to itself.
+			// Preserve the original parent from the existing entry instead.
+			if checkpoint.Parent == checkpoint.ID {
+				checkpoint.Parent = cp.Parent
+			}
 			// Merge ReferencedBy from the existing entry so we don't lose
 			// backup names accumulated by previous runs.
 			checkpoint.ReferencedBy = mergeReferencedBy(cp.ReferencedBy, checkpoint.ReferencedBy)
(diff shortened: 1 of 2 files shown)


pkg/uploader/run.go
```diff
@@ -640,6 +650,60 @@ var newKubeClient = func() (client.Client, error) {
...
+		}
+		checkpoints[idx].ReferencedBy = mergeReferencedBy(
+			checkpoints[idx].ReferencedBy, []string{backupName})
+		parentID = checkpoints[idx].Parent
+	}
+}
+
+// mergeReferencedBy returns the union of two ReferencedBy slices, preserving
+// order (existing entries first) and eliminating duplicates.
+func mergeReferencedBy(existing, additional []string) []string {
```
shubham-pampattiwar, line 699: mergeReferencedBy deduplicates entries from additional but trusts existing blindly. If the index data already has duplicates (corrupted file, older buggy write, etc.), they'll get carried forward on every merge. Cheapest fix is to run existing through the seen map too:
```
merged := make([]string, 0, len(existing)+len(additional))
for _, v := range existing {
    if !seen[v] {
        merged = append(merged, v)
        seen[v] = true
    }
}
```

pkg/uploader/run.go
```diff
@@ -356,7 +356,9 @@ func updateVMIndex(
 	found := false
 	for i, cp := range vmIndex.Checkpoints {
 		if cp.ID == checkpoint.ID {
```
shubham-pampattiwar, line 358: When a backup re-runs and produces a checkpoint with an ID that already exists in the index, we overwrite the old entry. If the new checkpoint's Parent ends up set to   
its own ID, we'd get a self-referential entry:                                                                                                                           
                                                                                                                                                                         
cp-002 (incr) → Parent: "cp-002"   // points to itself, restore chain is broken
```go
propagateReferencedBy survives this (has a visited guard), but buildCheckpointChain would loop. Simple guard before the overwrite:
if checkpoint.Parent == checkpoint.ID {
    checkpoint.Parent = cp.Parent
}
```

pkg/uploader/run_test.go
```diff
@@ -583,6 +583,540 @@ func countOccurrences(data []byte, substr string) int {
...
+				}
+
+				if len(idx.Checkpoints) != 1 {
+					t.Fatalf("expected 1 checkpoint, got %d", len(idx.Checkpoints))
+				}
+				assertReferencedBy(t, "cp-001", idx.Checkpoints[0].ReferencedBy, []string{"backup-day1"})
+			},
+		},
+		{
+			name: "dedup checkpoint merges referencedBy",
```
shubham-pampattiwar, line 944: The dedup test case covers a full backup re-run. Can we add one where an incremental backup re-runs with the same checkpoint ID? That's the case where the Parent field needs to survive the overwrite. e.g., cp-002 with Parent: "cp-001" gets overwritten, and we want to make sure Parent stays "cp-001" and doesn't get corrupted.

mpryc:
Tested with multiple backups, twice breaking chain, everything works as expected. Screenshot from the kubevirt-datamover-monitor app (from migtools), the referenced backups are showen there clearly:

image

Attached file [index.json](https://github.com/user-attachments/files/25609370/index.11.json) from those backups.

shubham-pampattiwar (approved):

sseago (approved):

PR #29, opened 2026-03-04
fix: serialize concurrent DataUploads per VM to prevent VMBT destruction

### LINKED ISSUE #28: BUG: Concurrent DataUploads for the same VM destroy each other's VMBTs
Summary
-------
When multiple Velero backups run concurrently for the same VirtualMachine, each DataUpload's reconcile loop deletes ALL VirtualMachineBackupTrackers (VMBTs) for that VM before creating its own. This destroys VMBTs that other DataUploads' VirtualMachineBackups (VMBs) are actively using, leaving those VMBs permanently stuck.

As of the latest fix https://github.com/migtools/kubevirt-datamover-controller/pull/27, stuck VMBs are now detected and the DataUpload is moved to Failed phase. The detection works by checking whether the VMB's referenced BackupTracker (VMBT) still exists in the cluster. If it was deleted, the DataUpload is failed immediately. If the VMBT still exists (e.g. PVC hotplug in progress), the controller keeps requeuing normally. However, the root cause — the indiscriminate VMBT deletion — is still present.

How to reproduce
----------------
1. Create a VM with CBT enabled (e.g. "cirros-test-cont-2" in namespace "cirros-test-2").
2. Trigger 3+ Velero backups in rapid succession (e.g. backup-1, backup-2, backup-3). Each backup creates a DataUpload for the same VM.
3. Observe that all but the last DataUpload fail with:
     "VMBackup stuck: BackupTracker vmbt-<name> no longer exists"

What happens step by step
-------------------------
Assume 3 DataUploads (DU-1, DU-2, DU-3) all target VM "my-vm" in namespace "vm-ns". The controller has MaxConcurrentReconciles=3 so they can run in parallel.

1. DU-1 enters handleAccepted:
   - Calls prepareVMBackupTracker("my-vm", "vm-ns")
   - No existing VMBTs found, creates VMBT-A (e.g. "vmbt-my-vm-abc12")
   - Annotates DU-1 with AnnotationVMBTName = "vmbt-my-vm-abc12"
   - Creates VMB-1 with spec.source.name = "vmbt-my-vm-abc12"

2. DU-2 enters handleAccepted (same time or moments later):
   - Calls prepareVMBackupTracker("my-vm", "vm-ns")
   - Lists ALL VMBTs with label LabelVMNameHash matching "my-vm"
   - Finds VMBT-A (created by DU-1) and DELETES it
   - Creates VMBT-B (e.g. "vmbt-my-vm-def34")
   - Creates VMB-2 with spec.source.name = "vmbt-my-vm-def34"

3. DU-3 enters handleAccepted:
   - Calls prepareVMBackupTracker("my-vm", "vm-ns")
   - Finds VMBT-B (created by DU-2) and DELETES it
   - Creates VMBT-C (e.g. "vmbt-my-vm-ghi56")
   - Creates VMB-3 with spec.source.name = "vmbt-my-vm-ghi56"

Result:
- Only VMBT-C exists.
- VMB-1 references VMBT-A (deleted) → stuck with "BackupTracker does not exist"
- VMB-2 references VMBT-B (deleted) → stuck with "BackupTracker does not exist"
- VMB-3 references VMBT-C (alive)   → may proceed, but see note below

Note: Even VMB-3 may not succeed. KubeVirt only allows one active (non-terminal) VMB per VirtualMachine. If VMB-1 or VMB-2 were already accepted by KubeVirt before their VMBTs were deleted, they block VMB-3 from progressing. KubeVirt does not automatically garbage-collect stuck VMBs.

Where the bug lives
--------------------
internal/controller/kubevirt_dataupload_controller.go

Function: prepareVMBackupTracker (line ~976)

The problematic code is the deletion loop at the top of the function:

    existingVMBTList := &kubevirtbackupv1alpha1.VirtualMachineBackupTrackerList{}
    r.List(ctx, existingVMBTList,
        client.InNamespace(vmNamespace),
        client.MatchingLabels{common.LabelVMNameHash: common.HashForLabel(vmName)})

    for i := range existingVMBTList.Items {
        r.Delete(ctx, &existingVMBTList.Items[i])   // <-- deletes ALL VMBTs for this VM
    }

This label selector (LabelVMNameHash) matches ALL VMBTs for the VM, regardless of which DataUpload created them. It does not check whether a VMBT is still referenced by an active VMB.

Why it was written this way
---------------------------
The design assumes one DataUpload per VM at a time. The intent is: "Delete whatever VMBT exists for this VM, then recreate it from S3 state."
This is correct for sequential backups (one at a time), but breaks when multiple DataUploads run concurrently for the same VM.

Existing guards (and why they don't help here)
----------------------------------------------
1. The "VMB already exists" guard in handleAccepted (line ~268):

       vmb, _ := r.findVMBForDataUpload(ctx, du, vmRef.Namespace)
       if vmb != nil {
           // skip VMBT preparation
       }

   This checks for a VMB belonging to THIS DataUpload (by LabelDataUploadUID).
   It does NOT check for VMBs belonging to OTHER DataUploads. So DU-2 doesn't
   see DU-1's VMB and proceeds to delete DU-1's VMBT.

2. The "AnnotationVMBTName" cache-lag guard (line ~276):

       if vmb == nil && du.Annotations[AnnotationVMBTName] != "" {
           return ctrl.Result{RequeueAfter: RequeueAfterShort}, nil
       }

   This protects a single DataUpload from its own rapid re-reconcile deleting
   its own VMBT. It does NOT protect against a DIFFERENT DataUpload deleting it.

3. KubeVirt admission webhook ("in progress for source"):

   KubeVirt blocks creating a second VMB for the same VM source. But each of our VMBs uses a different source (different VMBT), so the webhook does not block them. The webhook checks source identity, not the underlying VM.

Suggested fix
-------------
Option A: Skip VMBTs referenced by active VMBs (minimal change)

In prepareVMBackupTracker, before deleting a VMBT, check if any non-terminal VMB references it. If so, skip deletion:

    for i := range existingVMBTList.Items {
        vmbtToDelete := &existingVMBTList.Items[i]

        // Check if any VMB still references this VMBT
        vmbList := &kubevirtbackupv1alpha1.VirtualMachineBackupList{}
        r.List(ctx, vmbList, client.InNamespace(vmNamespace))
        referenced := false
        for _, vmb := range vmbList.Items {
            if vmb.Spec.Source.Name == vmbtToDelete.Name {
                referenced = true
                break
            }
        }
        if referenced {
            logger.Info("Skipping VMBT deletion, still referenced by VMB",
                "vmbt", vmbtToDelete.Name)
            continue
        }

        r.Delete(ctx, vmbtToDelete)
    }

Option B: Serialize backups per VM (stronger guarantee)

Before entering prepareVMBackupTracker, check if any OTHER DataUpload for the same VM already has a non-terminal VMB. If so, requeue with a delay:

    // In handleAccepted, before calling prepareVMBackupTracker:
    allVMBs := list all VMBs in vmNamespace with LabelVMNameHash matching this VM
    for _, vmb := range allVMBs {
        if vmb is not owned by this DU and vmb is not Done {
            logger.Info("Another backup is in progress for this VM, requeuing")
            return ctrl.Result{RequeueAfter: RequeueAfterLong}, nil
        }
    }

This ensures only one DataUpload at a time creates VMBT+VMB for a given VM.
The others wait their turn.

Option C: Combine both

Use Option B as the primary guard (serialize access), with Option A as a safety net (never delete in-use VMBTs).

Affected files
--------------
- internal/controller/kubevirt_dataupload_controller.go
  - prepareVMBackupTracker(): VMBT deletion loop (line ~977-989)
  - handleAccepted(): add per-VM serialization guard before VMBT preparation

Current mitigation
------------------
The controller now detects deleted VMBTs at two points in handleAccepted,
covering both cases observed in practice:

1. vmb.Status == nil (KubeVirt never started processing the VMB):
   Before requeuing with "status not yet available", the controller checks
   whether vmb.Spec.Source.Name (the VMBT) still exists. If it's gone, the
   DataUpload is failed immediately. This covers bkpc-2..9-style VMBs where
   KubeVirt never sets any conditions because the VMBT was already deleted.

2. vmb.Status has conditions (Initializing=True + Progressing=False):
   Inside evaluateVMBackupStatus, the controller fetches the VMBT referenced
   in vmb.Spec.Source.Name. If the VMBT is gone (NotFound), the DataUpload
   is failed. If the VMBT still exists, the controller treats it as a normal
   transient state (e.g. PVC being hotplugged to the VMI) and keeps requeuing.

Both checks produce the same error message:
  "VMBackup stuck: BackupTracker <name> no longer exists"

This prevents infinite silent requeuing, but the backup still fails rather than succeeding. The fix described above would prevent the failure from happening in the first place.

mpryc:
Fixes #28 

  ### Problem                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                         
When multiple Velero backups run concurrently for the same VM, each DataUpload's `prepareVMBackupTracker` deletes ALL VMBTs for that VM before creating its own. This destroys VMBTs that other DataUploads' VMBs are actively using, leaving those VMBs permanently stuck with `"BackupTracker no longer exists"`.                                                                                                                                                                                                                                                  

### Root cause

The VMBT deletion loop in `prepareVMBackupTracker` uses a label selector (`LabelVMNameHash`) that matches all VMBTs for a VM, regardless of which DataUpload owns them. With `MaxConcurrentReconciles=3`, multiple DataUploads can enter this function simultaneously and delete each other's VMBTs.

### Fix

Three layers, each targeting a different part of the problem:

**1. Serialize backups per VM (`handleAccepted`)**

A new `hasOlderActiveDUForVM` guard at the top of `handleAccepted` lists all DataUploads in the namespace and checks if an older one (by `CreationTimestamp`, then UID as tiebreaker) targeting the same VM is still in an active phase (`Accepted`, `Prepared`, or `InProgress`). If found, the current DataUpload requeues with a 30s delay. This ensures only one DataUpload at a time creates resources for a given VM, which also enables incremental backups — each backup completes its S3 upload before the next one starts, so the checkpoint chain is preserved.

**2. Skip referenced VMBTs (`prepareVMBackupTracker`)**

  Defense-in-depth for the TOCTOU race window where two reconcilers pass the serialization check simultaneously. Before deleting a VMBT, the deletion loop now pre-fetches all VMBs in the namespace and checks if any non-terminal VMB (using `isVMBTerminal` helper) still references that VMBT via `Spec.Source.Name`. If referenced, the VMBT is skipped with a log message instead of deleted.

  **3. Reconstruct VMBT in datamover pod (`pkg/uploader/run.go`)**

When the datamover pod (running in `InProgress` phase) tries to archive the VMBT to S3 but finds it was already deleted, it now reconstructs a minimal VMBT from the uploader config instead of failing. This is safe because the only field the controller reads from the archived `vmbt.json` is `Status.LatestCheckpoint`, which the pod always overwrites with the current checkpoint before archiving.

## Tests after fixes

Performed 2 tests:

### First was to create 15 backups all at once, all successfully finished

<img width="1886" height="880" alt="image" src="https://github.com/user-attachments/assets/13c4463a-5486-4fd4-a197-1f064e8abdc2" />

### Second test
Was similar to previous, but after first set of backups were done, another 20 backups were created all at once and in the middle of their work manual removal of one of the s3 backups was performed to ensure next one that was queued will be a full backup meaning that the chain breakage and follow up backups are properly discovering chains.

This test was also successful !
  
<img width="2175" height="950" alt="image" src="https://github.com/user-attachments/assets/e7a93a85-8ac1-45ef-803e-82fdd6dd193e" />




<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Per-VM serialization for DataUploads so newer uploads wait for older ones to finish, avoiding checkpoint conflicts.
  * Safer cleanup logic to avoid deleting backup trackers still referenced by active backups.

* **Bug Fixes**
  * Graceful reconstruction when a backup tracker is missing during archival, preventing failure.

* **Tests**
  * Large expansion of test coverage for sequencing, terminal-state detection, checkpoint lookup, archival, and edge cases.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/internal/controller/kubevirt_dataupload_controller.go
+++ b/internal/controller/kubevirt_dataupload_controller.go
@@ -234,6 +234,20 @@ func (r *KubeVirtDataUploadReconciler) handleAccepted(ctx context.Context, logge
 		return ctrl.Result{}, nil
 	}
 
+	// Serialize per VM: only the oldest DataUpload for a given VM proceeds.
+	// All younger DUs wait until the older one reaches a terminal phase (Completed/Failed).
+	// This ensures the previous backup's checkpoint is uploaded to S3 before the
+	// next backup starts, enabling incremental backups.
+	shouldWait, blockingDU, err := r.hasOlderActiveDUForVM(ctx, du)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if shouldWait {
+		logger.Info("Another DataUpload is still active for this VM, waiting",
+			"vm", vmRef.Name, "blockingDU", blockingDU)
+		return ctrl.Result{RequeueAfter: RequeueAfterLong}, nil
+	}
+
 	// Step 0: Verify BSL exists and is Available before creating any resources.
 	// BSL must be in Available phase — if it's not, fail immediately rather than
 	// creating resources (PVC, VMBT, VMB) that will be wasted.
@@ -999,16 +1013,40 @@ func (r *KubeVirtDataUploadReconciler) ensureTempPVC(ctx context.Context, logger
 
 // prepareVMBackupTracker creates a VirtualMachineBackupTracker for the VM,
 // restoring the LatestCheckpoint from the archived VMBT in S3 if available.
-// Any existing VMBT is deleted first to avoid conflicts (S3 is the source of truth).
+// Existing VMBTs are deleted first unless they are still referenced by an active
+// (non-terminal) VMB, to avoid destroying VMBTs used by concurrent DataUploads.
 func (r *KubeVirtDataUploadReconciler) prepareVMBackupTracker(ctx context.Context, logger logr.Logger, du *velerov2alpha1.DataUpload, vmName, vmNamespace string) (*kubevirtbackupv1alpha1.VirtualMachineBackupTracker, error) {
-	// Delete any existing VMBT for this VM before creating a new one.
-	// This ensures we always start from the state restored from S3.
+	// Delete existing VMBTs for this VM before creating a new one, but only if
+	// they are not still referenced by an active (non-terminal) VMB. This prevents
+	// concurrent DataUploads from destroying each other's VMBTs.
 	existingVMBTList := &kubevirtbackupv1alpha1.VirtualMachineBackupTrackerList{}
 	if err := r.List(ctx, existingVMBTList, client.InNamespace(vmNamespace), client.MatchingLabels{common.LabelVMNameHash: common.HashForLabel(vmName)}); err != nil {
 		return nil, fmt.Errorf("failed to list existing VMBTs: %w", err)
 	}
+
+	// Pre-fetch all VMBs in the namespace to check references
+	allVMBs := &kubevirtbackupv1alpha1.VirtualMachineBackupList{}
+	if err := r.List(ctx, allVMBs, client.InNamespace(vmNamespace)); err != nil {
+		return nil, fmt.Errorf("failed to list VMBs: %w", err)
+	}
+
 	for i := range existingVMBTList.Items {
 		vmbtToDelete := &existingVMBTList.Items[i]
+
+		// Check if any non-terminal VMB still references this VMBT
+		referenced := false
+		for j := range allVMBs.Items {
+			if allVMBs.Items[j].Spec.Source.Name == vmbtToDelete.Name && !isVMBTerminal(&allVMBs.Items[j]) {
+				referenced = true
+				break
+			}
+		}
+		if referenced {
+			logger.Info("Skipping VMBT deletion, still referenced by active VMB",
+				"vmbt", vmbtToDelete.Name)
+			continue
+		}
+
 		logger.Info("Deleting existing VMBT before recreation", "vmbt", vmbtToDelete.Name)
 		if err := r.Delete(ctx, vmbtToDelete); err != nil && !errors.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to delete existing VMBT %s: %w", vmbtToDelete.Name, err)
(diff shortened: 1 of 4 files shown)


sseago (approved):

shubham-pampattiwar (commented):
Minor observation (not a blocker for this PR): if an older DataUpload gets stuck in an active phase indefinitely (e.g., controller crash mid-reconcile), younger DUs for the same VM will keep requeuing at 30s intervals forever. I believe Velero's existing DataUpload timeout should eventually move the stuck DU to a terminal phase, but worth confirming. If not, we may want to add a staleness check in a follow-up — e.g., skip a blocking DU if it's been in Accepted/Prepared for longer than some threshold.

shubham-pampattiwar (approved):

PR #54, opened 2026-04-22
Issue 46: Add pvcSizes to per-VM index file

sseago:
This PR adds pvcSizes to the per-VM index file so on restore we won't need an additional APIServer call to know what size to create the PVC where we're copying the reconstructed filesystem.

In addition, this PR fixes a bug where we were listing the PVC name in the index as "volume0" or "volume1" rather than using the actual PVC name.

## Summary by CodeRabbit

* **New Features**
  * Checkpoint metadata now includes PVC storage sizes for improved capacity visibility.

* **Improvements**
  * Image pull policy now defaults to Always when not configured, ensuring fresh datamover image pulls.
  * Stricter volume validation and a new VM volume→PVC mapping to more reliably resolve and report PVC references.

* **Tests**
  * Added unit tests covering VM volume→PVC mapping and uploader/index behavior with simulated Kubernetes objects.

### DIFF
--- a/internal/controller/kubevirt_dataupload_controller.go
+++ b/internal/controller/kubevirt_dataupload_controller.go
@@ -1377,7 +1377,7 @@ func (r *KubeVirtDataUploadReconciler) buildDatamoverPodConfig(
 
 	pullPolicy := r.DatamoverImagePullPolicy
 	if pullPolicy == "" {
-		pullPolicy = corev1.PullIfNotPresent
+		pullPolicy = corev1.PullAlways
 	}
 
 	return &DatamoverPodConfig{
--- a/pkg/uploader/run.go
+++ b/pkg/uploader/run.go
@@ -33,6 +33,7 @@ import (
 	velero "github.com/vmware-tanzu/velero/pkg/plugin/velero"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -109,7 +110,7 @@ func Run(ctx context.Context, logger logr.Logger) error {
 	logger.Info("Kube resources archived to S3")
 
 	// Update VM index (references the archived paths)
-	if err := updateVMIndex(ctx, store, config, files, archived, logger); err != nil {
+	if err := updateVMIndex(ctx, store, k8sClient, config, files, archived, logger); err != nil {
 		return fmt.Errorf("failed to update VM index: %w", err)
 	}
 
@@ -277,7 +278,7 @@ func extractDiskName(filename string) string {
 // that were uploaded by archiveKubeResources — these are stored on the
 // checkpoint entry so the index always references valid objects.
 func updateVMIndex(
-	ctx context.Context, store velero.ObjectStore, config *UploaderConfig,
+	ctx context.Context, store velero.ObjectStore, k8sClient client.Client, config *UploaderConfig,
 	files []CheckpointFile, archived *archivedPaths, logger logr.Logger,
 ) error {
 	indexPath := fmt.Sprintf("checkpoints/%s/%s/index.json", config.VMNamespace, config.VMName)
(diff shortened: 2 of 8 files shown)


pkg/uploader/run.go
```diff
@@ -310,11 +313,32 @@ func updateVMIndex(
...
+			pvcName := volumeMap[f.DiskName]
+			if pvcName == "" {
+				pvcName = f.DiskName
+			}
+			pvcNames = append(pvcNames, pvcName)
+			pvc := &corev1.PersistentVolumeClaim{}
+			if err := k8sClient.Get(ctx, types.NamespacedName{Name: pvcName, Namespace: config.VMNamespace}, pvc); err != nil {
+				return fmt.Errorf("failed to get PVC %s: %w", pvcName, err)
+			}
+			if storage, ok := pvc.Spec.Resources.Requests[corev1.ResourceStorage]; ok {
```
shubham-pampattiwar, line 340: If any PVC is missing a storage request, `PVCs[i]` and `PVCSizes[i]` will be misaligned. Either always append (use zero `resource.Quantity{}` as fallback) or return an error.

pkg/uploader/run_test.go
```diff
@@ -555,7 +561,28 @@ func TestUpdateVMIndex(t *testing.T) {
...
+					})
+				}
+			}
+			objects = append(objects, &kubevirtcorev1.VirtualMachine{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-vm",
+					Namespace: "test-ns",
+				},
+			})
+			fakeClient := fake.NewClientBuilder().
```
shubham-pampattiwar, line 616: The lint job is failing due to gocyclo on TestUpdateVMIndex_ReferencedByPropagation (complexity 32, limit 30). The repeated fake client setup block is duplicated across
  three test functions, extracting a helper would fix it:                                                                                                                 
                  
```go
  func buildTestClient(t *testing.T, scheme *runtime.Scheme, config *UploaderConfig, files []CheckpointFile) client.Client {                                               
        t.Helper()                                                                                                                                                         
        var objects []client.Object
        for _, f := range files {                                                                                                                                          
                if f.DiskName != "" {
                        objects = append(objects, &corev1.PersistentVolumeClaim{
                                ObjectMeta: metav1.ObjectMeta{
                                        Name:      f.DiskName,
                                        Namespace: config.VMNamespace,
                                },
                        })
                }
        }
        objects = append(objects, &kubevirtcorev1.VirtualMachine{
                ObjectMeta: metav1.ObjectMeta{
                        Name:      config.VMName,
                        Namespace: config.VMNamespace,
                },
        })
        return fake.NewClientBuilder().
                WithScheme(scheme).
                WithObjects(objects...).Build()
  }
```

pkg/uploader/run_test.go
```diff
@@ -1097,7 +1128,28 @@ func TestUpdateVMIndex_ReferencedByPropagation(t *testing.T) {
 				}
 			}
 
-			err := updateVMIndex(context.Background(), store, tt.config, tt.files, tt.archived, logr.Discard())
+			var objects []client.Object
+			for _, f := range tt.files {
+				if f.DiskName != "" {
+					objects = append(objects, &corev1.PersistentVolumeClaim{
```
shubham-pampattiwar: The test PVC objects are created without `Spec.Resources.Requests`, so PVCSizes is always empty in the updateVMIndex test results, the function isn't actually run end-to-end.

pkg/uploader/run.go
```diff
@@ -310,11 +313,34 @@ func updateVMIndex(
...
+	volumeMap := common.GetVolumeMapForVm(vm)
+
 	// Extract PVC/disk names from uploaded files
 	var pvcNames []string
+	var pvcSizes []resource.Quantity
 	for _, f := range files {
 		if f.DiskName != "" {
-			pvcNames = append(pvcNames, f.DiskName)
+			pvcName := volumeMap[f.DiskName]
+			if pvcName == "" {
```
shubham-pampattiwar, line 331: @sseago When `volumeMap[f.DiskName]` is empty, falling back to `f.DiskName` will use the extracted qcow2 disk name (e.g. "disk1"), which is unlikely to match an actual PVC name. Since this PR is specifically fixing PVC name resolution, should we fail fast here instead of masking a mapping failure?

shubham-pampattiwar (commented):
just one last question

pkg/uploader/run.go
sseago: Oh. Actually, probably yes. Let me make that last change.

shubham-pampattiwar (approved):
/lgtm

internal/controller/kubevirt_dataupload_controller.go
```diff
@@ -1377,7 +1377,7 @@ func (r *KubeVirtDataUploadReconciler) buildDatamoverPodConfig(
 
 	pullPolicy := r.DatamoverImagePullPolicy
 	if pullPolicy == "" {
-		pullPolicy = corev1.PullIfNotPresent
+		pullPolicy = corev1.PullAlways
 	}
```
kaovilai, line 1381: shouldnt this follow operator to default to Ifnotpresent on shas aka downstream builds etc. https://github.com/openshift/oadp-operator/blob/5427567508f1ae23a93694da231627ae8714cf13/docs/design/tech-debt/image-pull-policy-error-handling.md?plain=1#L82

sseago: Although maybe we shouldn't duplicate that logic here. OADP is already doing that to set the pull policy for the controller deployment. It could set an env var as well, and then this controller would use the env var, falling back to Always. If we do that, then we probably want the above line as-is, then once we get the env var set up from oadp-operator, we can submit a follow-on PR here to use the env var if present.

mpryc (approved):

Issue #73, opened 2026-05-12
Implement DataDownload controller for VM restore


shubham-pampattiwar:
## Problem

The kubevirt-datamover-controller currently only supports the backup (upload) path via the DataUpload controller. There is no restore path -- no DataDownload controller exists to download checkpoint data from S3 and rebuild VM disks during a Velero restore.

Related issues:
- #44 -- BIA changes needed for restore
- #45 -- `qemu-img` binary needed for qcow2 chain reconstruction

## Proposed Approach

Implement a DataDownload controller that mirrors the DataUpload flow but in reverse:

1. Velero creates a DataDownload CR during restore
2. Controller reads the per-backup-per-VM manifest from S3 to get the checkpoint chain
3. Downloads qcow2 files for each checkpoint in the chain (full + incrementals) from S3
4. Uses `qemu-img` to reconstruct the disk by applying the incremental chain on top of the full backup
5. Provisions a new PVC with the reconstructed disk data
6. Marks DataDownload as complete so Velero can continue the restore

Key design decisions to resolve:
- **Disk reconstruction strategy**: Download all qcow2 files in the chain and use `qemu-img rebase`/`commit` to flatten, or stream and merge on the fly
- **Temporary storage**: The downloader pod needs scratch space to hold the chain during reconstruction -- sizing and storage class considerations
- **RIA plugin**: A RestoreItemAction plugin is needed to create the DataDownload CR and coordinate with Velero's restore flow
- **PVC provisioning**: How to create the target PVC with correct size, storage class, and access mode (from `pvcSizes` in the index, per #46)

## Files to add/change

- `internal/controller/kubevirt_datadownload_controller.go` -- new controller watching DataDownload CRs
- `pkg/downloader/` -- new package for S3 download and qcow2 chain reconstruction logic
- `internal/controller/pod_builder.go` -- downloader pod spec (needs `qemu-img`, scratch storage, S3 credentials)
- `cmd/main.go` -- register the DataDownload controller
- RIA plugin (may live in a separate repo or alongside the DIA plugin)

## Acceptance Criteria

- [ ] DataDownload controller watches and reconciles DataDownload CRs
- [ ] Downloads full checkpoint chain from S3 for a given VM
- [ ] Reconstructs disk using `qemu-img` from the incremental chain
- [ ] Provisions a PVC with the restored disk data
- [ ] Handles single-disk and multi-disk VMs
- [ ] Marks DataDownload as complete/failed with appropriate status messages
[...]

PR #115, opened 2026-07-16
feat: implement downloader runtime for VM restore (issue #73 Phase 2)

kaovilai:
## Summary

Implements Phase 2 of the phased plan on #73: the downloader's core data path for VM restore, as a standalone `pkg/downloader` package.

- `pkg/downloader/chain.go` — `resolveCheckpointFiles`: walks the already-resolved checkpoint chain (`VMBackupManifest.CheckpointChain`) and filters each checkpoint's files down to the one matching the target disk/volume.
- `pkg/downloader/download.go` — `downloadCheckpointFiles`: streams each checkpoint's qcow2 file from object storage to local scratch storage, in chain order.
- `pkg/downloader/reconstruct.go` — `rebaseChain`/`flattenToRaw`: wraps `qemu-img rebase`/`convert` to repoint backing-file references onto local scratch paths and flatten the chain into a single raw disk image.
- `pkg/downloader/run.go` — `Run`/`LoadConfigFromEnv`: orchestrates the full restore data path, mirroring `pkg/uploader/run.go`'s shape.
- `pkg/downloader/types.go` — `DownloaderConfig` and its `KUBEVIRT_DM_*` env vars.

Also extracts `common.ObjectStoreConfig` out of `uploader.UploaderConfig` (pure refactor, no behavior change) so the downloader can reuse `InitObjectStore`/`GetVMBackupManifest`/`GetVMIndex` without depending on the uploader's full config shape.

**Out of scope for this PR** (deferred to Phase 3, per #83's own description and the issue's phased plan):
- `internal/controller/kubevirt_datadownload_controller.go` (the actual reconciler)
- `cmd/main.go` `download` subcommand dispatch
- Downloader pod builder wiring
- RBAC, RIA plugin, multi-disk hardening, e2e tests

## Verification

- `go build ./...`, `go vet ./...`, `make test`, `make lint` all clean.
- Manually verified the real data path end-to-end against production S3 checkpoint data (`s3://tkaovila-oadp/kubevirt-dm-amd64-kubevirt-datamover/`, `cirros-test` full backup): `GetVMIndex` → `resolveCheckpointFiles` → `downloadCheckpointFiles` → `rebaseChain` → `flattenToRaw` produced a raw disk image whose size matched the source qcow2's virtual size exactly (1073741824 bytes).
- Ran 15 rounds of `coderabbit review --agent`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test`
- [x] `make lint`
- [x] Manual smoke test against real S3 checkpoint data (see above)
- [x] 15 rounds of `coderabbit review --agent`
- [ ] CI

> [!Note]
> Responses generated with Claude

## Summary by CodeRabbit

* **New Features**
  * Added automated VM disk restoration from backup checkpoints, including ordered checkpoint-chain resolution, checkpoint downloading, QCOW2 rebasing/flattening, and atomic publishing of the restored raw disk image.
  * Introduced environment-based restore configuration (target volume/path, scratch staging, and credentials).
  * Added shared object-store configuration support used consistently across backup and restore.

* **Bug Fixes**
  * Strengthened validation for missing/duplicated/invalid checkpoint data and safer filename handling to prevent path traversal.
  * Added robust cleanup on failures and protections against partially-written outputs (including replacing pre-existing symlinks).


pkg/downloader/download.go
```diff
@@ -0,0 +1,96 @@
...
+		_ = os.Remove(localPath)
+		return fmt.Errorf("failed to write local file %q: %w", localPath, err)
+	}
+	if err := out.Close(); err != nil {
+		_ = os.Remove(localPath)
+		return fmt.Errorf("failed to finalize local file %q: %w", localPath, err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
kaovilai: Good catch — fixed in 5b52b99. `downloadOne` now captures the reader's `Close()` error alongside the local file's, instead of discarding it via a deferred call, and cleans up the partial local file if either fails. Added a test that simulates a reader whose `Close()` errors to cover this.

> [!Note]
> Responses generated with Claude

pkg/downloader/chain.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"fmt"
+
+	"github.com/migtools/kubevirt-datamover-controller/pkg/uploader"
+)
+
+// resolveCheckpointFiles walks the (already-ordered) checkpoint chain and
+// returns, for each checkpoint, the single CheckpointFile matching
+// targetVolume. The chain is assumed to be full-backup-first,
+// incrementals-after, as produced by the uploader at backup time.
```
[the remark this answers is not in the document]
kaovilai: Fixed in ff408d3 — `resolveCheckpointFiles` now validates each entry's `Type`/`Parent` against its position in the chain (first entry must be full with no parent, later entries must be incremental with `Parent == chain[i-1]`). Added reversed-chain and disconnected-chain tests.

> [!Note]
> Responses generated with Claude

pkg/downloader/download.go
```diff
@@ -0,0 +1,127 @@
...
+	// Width the index prefix to the chain length so lexical and chain order
+	// agree even past 100 checkpoints (e.g. "005-" sorts before "100-").
+	indexWidth := max(len(fmt.Sprintf("%d", len(files)-1)), 2)
+
+	localPaths := make([]string, 0, len(files))
+	cleanup := func() {
+		for _, p := range localPaths {
+			_ = os.Remove(p)
+		}
+	}
```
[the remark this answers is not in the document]
kaovilai: Fixed in ff408d3 — `cleanup` now returns any `os.Remove` failures (joined with the triggering cancellation/download error via `errors.Join`) instead of discarding them.

> [!Note]
> Responses generated with Claude

pkg/downloader/download.go
```diff
@@ -0,0 +1,127 @@
...
+			cleanup()
+			return nil, err
+		}
+		if file.Filename == "" || file.Filename != filepath.Base(file.Filename) ||
+			file.Filename == "." || file.Filename == ".." {
+			cleanup()
+			return nil, fmt.Errorf("invalid checkpoint filename %q", file.Filename)
+		}
+		localPath := filepath.Join(scratchDir, fmt.Sprintf("%0*d-%s", indexWidth, i, file.Filename))
+		if err := downloadOne(store, bucket, file.ObjectPath, localPath); err != nil {
```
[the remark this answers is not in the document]
kaovilai: Fixed in ff408d3 — `downloadOne` now takes the checkpoint's declared `Size`, streams through `io.LimitReader(reader, size+1)`, and rejects the download if the copied byte count doesn't exactly match (catches both undersized and oversized/replaced objects). Added tests for both cases. Also verified against real production S3 data that the declared size already matches the actual object's `ContentLength`, so this doesn't reject genuine checkpoints.

> [!Note]
> Responses generated with Claude

pkg/downloader/reconstruct.go
```diff
@@ -0,0 +1,148 @@
...
+		Data struct {
+			DataFile string `json:"data-file"`
+		} `json:"data"`
+	} `json:"format-specific"`
+}
+
+func inspectQcow2(ctx context.Context, path string) (qcow2Info, error) {
+	stdout, stderr, err := runQemuImg(ctx, "info", "-f", "qcow2", "--output=json", path)
+	if err != nil {
+		return qcow2Info{}, fmt.Errorf("failed to inspect %q: %w (%s)", path, err, stderr)
```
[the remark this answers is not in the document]
kaovilai: Checked this empirically rather than relying on the general web research cited above, since it makes a specific factual claim about this exact invocation. Created a qcow2 whose `backing-filename` points at a FIFO (which blocks on open) and ran the exact command this code uses: `qemu-img info -f qcow2 --output=json <path>` (no `--backing-chain`). It returned instantly without touching the FIFO — confirming this qemu-img version does not open/dereference the backing file for a plain `info` call; only the explicit `--backing-chain` flag (never used in this code) triggers that traversal.

Also checked `qemu-img info --help`: `-U`/`--force-share` is documented as "open image in shared mode for concurrent access" (locking behavior), not backing-chain traversal control — so it wouldn't have been the correct fix even if the premise held.

Given this, I don't believe there's a real gap here for this code's actual invocation pattern. Leaving this thread open in case a maintainer wants to double check on a different qemu-img version, but not changing the code based on the general (version-unspecified) web research cited in the analysis above.

> [!Note]
> Responses generated with Claude

kaovilai:
@coderabbitai full review

pkg/downloader/run.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	files, err := resolveCheckpointFiles(
+		config.VMNamespace, config.VMName, manifest.CheckpointChain, vmIndex, config.TargetVolume)
+	if err != nil {
+		return fmt.Errorf("failed to resolve checkpoint chain: %w", err)
+	}
+
+	logger.Info("Resolved checkpoint chain", "checkpoints", len(files))
+	if len(files) == 0 {
```
shubham-pampattiwar: nit: This check is unreachable, `resolveCheckpointFiles` returns one file per chain entry, and the chain was already validated non-empty by the manifest check above. If `resolveCheckpointFiles` returns nil error, `len(files) == len(chain) > 0` always holds

pkg/downloader/reconstruct.go
```diff
@@ -0,0 +1,151 @@
...
+		_ = f.Close()
+		_ = os.Remove(outputPath)
+		return fmt.Errorf("failed to pre-create raw image %q: %w", outputPath, err)
+	}
+	if err := f.Close(); err != nil {
+		_ = os.Remove(outputPath)
+		return fmt.Errorf("failed to pre-create raw image %q: %w", outputPath, err)
+	}
+
+	_, stderr, err := runQemuImg(ctx, "convert", "-f", "qcow2", "-O", "raw", chainTipPath, outputPath)
```
shubham-pampattiwar: For large VMs, this convert step could take minutes with no log output. Consider adding -p to get progress percentages in stderr, helpful for operators watching pod logs during restore.

kaovilai: Good catch — confirmed empirically that `-p` writes progress to qemu-img's stdout, not stderr. Fixed in 0336bb1: added `-p` to the convert invocation, and changed `runQemuImg`'s real implementation to mirror stdout/stderr live to the process's own stdout/stderr (in addition to the existing buffered capture used for error messages), so the progress output actually reaches `kubectl logs` instead of being silently discarded on the success path.

> [!Note]
> Responses generated with Claude

pkg/downloader/types.go
```diff
@@ -0,0 +1,105 @@
...
+	EnvDataDownloadName = "KUBEVIRT_DM_DATADOWNLOAD_NAME"
+	EnvDataDownloadUID  = "KUBEVIRT_DM_DATADOWNLOAD_UID"
+	EnvTargetVolume     = "KUBEVIRT_DM_TARGET_VOLUME"
+	EnvTargetPath       = "KUBEVIRT_DM_TARGET_PATH"
+	EnvScratchPath      = "KUBEVIRT_DM_SCRATCH_PATH"
+
+	// BSL/credentials env vars are shared with the uploader — upload and
+	// download pods never run in the same process, so there's no collision.
+	// See pkg/common.Env* for the canonical definitions.
+	EnvBSLProvider              = common.EnvBSLProvider
```
sseago: Is there any need to duplicate these instead of just referencing the common.* consts directly?

kaovilai: Addressing.

kaovilai: if merged first will dedupe in followup

kaovilai: Good point — done in aa84edf. Removed the re-exported `Env*`/`DefaultCredentialsPath` aliases from both `pkg/downloader/types.go` and `pkg/uploader/types.go` entirely; call sites in both packages' `run.go`/tests now reference `common.Env*`/`common.DefaultCredentialsPath` directly. Also updated `internal/controller/pod_builder.go` (and its test), which was referencing the uploader aliases externally, to reference `common.*` directly as well — so there's now exactly one place (`pkg/common/env.go`) that owns these names, and adding a new one doesn't require touching two more files to re-export it.

> [!Note]
> Responses generated with Claude

PR #124, opened 2026-07-17
feat: implement DataDownload controller for VM restore (issue #73 Phase 3)

kaovilai:
## Summary

Implements Phase 3 of the phased plan on #73: the `DataDownload` controller that drives the restore data path (`New` → `Accepted` → `Prepared` → `InProgress` → `Completed`/`Failed`/`Canceling`), mirroring the existing `DataUpload` controller's shape.

**Depends on #115 (Phase 2, downloader runtime)**, which hasn't merged yet — this branch is based on it, so the diff currently includes #115's commits until that lands. Once #115 merges, this PR's diff will shrink to just Phase 3's changes.

- `internal/controller/kubevirt_datadownload_controller.go` (new): full phase state machine, `Spec.Cancel` handling, idempotent resume after partial completion (detects an already-committed rebind via the PV's `claimRef` + UID label rather than the target PVC's own `Status.Phase`, which the Kubernetes PV controller only updates asynchronously), scratch PVC sizing derived from checkpoint chain metadata (floored by the target PVC's own requested capacity).
- `internal/controller/pv_rebind.go`: adds `BindTargetMode` (`BindTargetCreate` for the upload path vs. `BindTargetExisting` for download) to the shared `rebindPVToNamespace` helper, plus `bindExistingPVC` compatibility validation (StorageClassName, capacity, VolumeMode, AccessModes, conflicting `Spec.VolumeName`) for binding a PV to a PVC Velero already created, instead of creating a new one.
- `internal/controller/pod_builder.go`: download-mode env vars/volumes for the downloader pod; the restored disk is written to `disk.img` at the scratch volume's root, matching the CDI/KubeVirt filesystem-mode PVC convention (this PV is rebound directly as the restore target's volume).
- `pkg/common/validation.go`: `GetVMReferenceFromDataDownload`, mirroring the existing DataUpload accessor.
- `pkg/downloader/chain.go`: exports `ResolveCheckpointFiles` (pure rename) so the controller's Accepted-phase chain validation/sizing can reuse it.
- `cmd/main.go`, `config/rbac/role.yaml`: `download` subcommand dispatch, controller registration, additive RBAC (`datadownloads`/`datadownloads/status`).

A real correctness bug caught along the way: the controller originally passed `dd.Spec.TargetVolume.PVC` (a PVC/claim name) directly to `ResolveCheckpointFiles`, which actually matches on `CheckpointFile.DiskName` (the KubeVirt volume name) — these differ whenever a VM's volume reference doesn't literally match its backing PVC's name. Fixed via `resolveTargetDiskName`, which translates PVC name → disk name using the backup's own index, persisted across reconciles via an annotation (mirroring the existing `AnnotationVMBTName` pattern).

## Review history

Reviewed over 5 rounds of `coderabbit --agent review --base` (against #115's branch, so only Phase 3's own diff was in scope) plus an independent reviewer subagent — fixed roughly 25 findings, including:
- PV-rebind validate-before-mutate ordering (fail fast on an incompatible target before tearing down the source PVC's reclaim policy)
- The PVC-name-vs-disk-name mismatch above
- Idempotent resume after a rebind succeeds but the phase update fails to persist
- Scratch-PVC sizing formula (target-PVC-capacity floor applied before overhead, not clamped after)
- The `disk.img` path convention fix

A handful of findings were left unfixed because they're pre-existing characteristics shared with the already-merged upload controller, not regressions introduced here — filed as follow-ups for discussion rather than diverging behavior between the two controllers: #120, #121, #122, #123.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (including `-race`)
- [x] `make lint`
- [ ] CI
- [ ] Full e2e restore validation (deferred to a later phase per the issue's own phased plan — this phase's exit criteria is phase-transition + pod-lifecycle unit tests)

> [!Note]
> Responses generated with Claude


internal/controller/kubevirt_datadownload_controller.go
```diff
@@ -0,0 +1,1905 @@
...
+// output PVC (see DatamoverPodConfig.OutputPVCName's doc comment).
+func (r *KubeVirtDataDownloadReconciler) buildDownloaderPodConfig(
+	dd *velerov2alpha1.DataDownload,
+	bsl *velerov1.BackupStorageLocation,
+	vmRef *common.VMReference,
+	scratchPVCName string,
+	outputPVCName string,
+	targetDiskName string,
+) (*DatamoverPodConfig, error) {
+	cfg, err := uploader.ExtractBSLConfig(bsl)
```
Joeavaikath, line 1336: buildDownloaderPodConfig does not propagate BSLProfile from the cfg
Just ensuring is this intentional?
Upload does use it
Upload controller (kubevirt_dataupload_controller.go:1934):
BSLProfile: cfg.Profile,

kaovilai: Good catch — confirmed, and it's broader than just `BSLProfile`: `buildDownloaderPodConfig` is also missing `BSLServerSideEncryption`, `BSLKMSKeyID`, `BSLKMSKeyName`, `BSLChecksumAlgorithm`, and the SSE-C secret fields (`SSECSecretName`/`SSECSecretKey`) that `buildUploadPodConfig` sets. Not intentional — tracked as a followup rather than fixing in this already-approved PR: #184

> [!Note]
> Responses generated with Claude

